PR #4, opened 2023-05-16
SDIT-405: ✨ Initial framework for clients

petergphillips:


build.gradle.kts
```diff
@@ -4,13 +4,32 @@ plugins {
...
 
 configurations {
+  implementation { exclude(module = "spring-boot-starter-web") }
+  implementation { exclude(module = "spring-boot-starter-tomcat") }
   testImplementation { exclude(group = "org.junit.vintage") }
 }
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-actuator")
+//  implementation("org.opensearch.client:spring-data-opensearch-starter:1.1.0")
```
petergphillips, line 14: excluded for now as PR was getting too big

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/config/WebClientConfiguration.kt
```diff
@@ -0,0 +1,86 @@
...
+@EnableAsync
+class WebClientConfiguration(
+  @Value("\${api.base.url.prison-api}") val prisonApiBaseUri: String,
+  @Value("\${api.base.url.incentives}") val incentivesBaseUri: String,
+  @Value("\${api.base.url.restricted-patients}") val restrictedPatientBaseUrl: String,
+  @Value("\${api.base.url.oauth}") val hmppsAuthBaseUri: String,
+) {
+
+  @Bean
+  fun hmppsAuthHealthWebClient(): WebClient = WebClient.builder().baseUrl(hmppsAuthBaseUri).build()
```
mikehalmamoj, line 25: If you inject the WebClient.Builder into this method then it comes wired up with the pre-configured Jackson object mapper.

mikehalmamoj: (although that is more pertinent to the non-health web clients which don't really need it!)

steverendell (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/config/WebClientConfiguration.kt
petergphillips: agreed, which is the approach I've taken for the non-health web clients

PR #5, opened 2023-05-17
SDIT-405: ✨ Integrate opensearch

petergphillips:


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 parameters:
   alerts-slack-channel:
     type: string
-    default: dps_alerts_security
+    default: syscon-alerts
```
petergphillips, line 9: was going to wrong channel

.circleci/config.yml
```diff
@@ -106,6 +106,7 @@ workflows:
                 - main
     jobs:
       - hmpps/gradle_owasp_dependency_check:
+          jdk_tag: "19.0"
```
petergphillips, line 109: owasp check was failing due to wrong jdk

docker-compose-test.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+services:
+  opensearch:
+    container_name: opensearch
+    image: opensearchproject/opensearch:2.5.0
```
petergphillips, line 5: might switch to localstack, but starting with opensearch for now

PR #6, opened 2023-05-17
SDIT-405: ✨ Add index status

petergphillips:


docker-compose-test.yml
```diff
@@ -1,19 +1,17 @@
...
-      - discovery.type=single-node
-      - bootstrap.memory_lock=true
-      - plugins.security.disabled=true
-      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
+    container_name: localstack-psi
     ports:
-      - "9200:9200"
+      - "4566:4566"
+    environment:
+      - SERVICES=opensearch
```
petergphillips, line 11: switch to localstack so can then include sqs and sns later easily

docker-compose-test.yml
```diff
@@ -1,19 +1,17 @@
...
+    volumes:
+      - "$PWD/src/test/resources/localstack/setup-opensearch.sh:/etc/localstack/init/ready.d/init-aws.sh"
```
petergphillips, line 13: not sure when this became the *new* way, but had to do this to get it working rather than mounting `docker-entrypoint-initaws.d`

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/repository/IndexStatusRepository.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.prisonersearchindexer.repository
+
+import org.springframework.data.elasticsearch.repository.ElasticsearchRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.prisonersearchindexer.model.IndexStatus
+
+@Repository
+interface IndexStatusRepository : ElasticsearchRepository<IndexStatus, String>
```
petergphillips, line 8: odd that it is still an elasticsearch repo rather than an opensearch repo

mikehalmamoj: Surprised at how much smaller the [spring data opensearch](https://github.com/opensearch-project/spring-data-opensearch/tree/main/spring-data-opensearch/src/main/java/org/opensearch/data/client/orhlc) library is. I guess it sits on top of [spring-data-elasticsearch](https://github.com/spring-projects/spring-data-elasticsearch/tree/main/src/main/java/org/springframework/data/elasticsearch)?

PR #11, opened 2023-05-23
SDIT-822: ⬆️ Upgrade to latest

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/HmppsPrisonerSearchIndexer.kt
```diff
@@ -2,9 +2,10 @@ package uk.gov.justice.digital.hmpps.prisonersearchindexer
 
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.autoconfigure.data.elasticsearch.ElasticsearchDataAutoConfiguration
+import org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchRestClientAutoConfiguration
 import org.springframework.boot.runApplication
 
-@SpringBootApplication(exclude = [ElasticsearchDataAutoConfiguration::class])
+@SpringBootApplication(exclude = [ElasticsearchDataAutoConfiguration::class, ElasticsearchRestClientAutoConfiguration::class])
```
petergphillips, line 8: not sure why this is now needed, but it is autowiring the elastic one instead of opensearch one without it

PR #16, opened 2023-06-05
SDIT-405: ✅ Start adding tests from probation indexer

petergphillips:


src/main/resources/application-dev.yml
```diff
@@ -35,8 +35,7 @@ spring:
           public-key-location: classpath:local-public-key.pub
 
 opensearch:
-  uris: http://localhost:4566
-  pathPrefix: /opensearch/eu-west-2/os01
+  uris: http://os01.eu-west-2.opensearch.localhost.localstack.cloud:4566
```
petergphillips, line 38: for circleci it is easier if we just have the url.  For running the indexer in docker we use both.

mikehalmamoj (approved):

PR #27, opened 2023-06-20
SDIT-405: ♻️ Switch back to non reactive

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/services/IndexService.kt
```diff
@@ -159,15 +160,17 @@ class IndexService(
...
       .also { logIndexStatuses(it) }
       .failIf(IndexStatus::isNotBuilding) { BuildNotInProgressError(it) }
       .map { doCancelIndexing() }
 
-  private suspend fun doCancelIndexing(): IndexStatus {
+  private fun doCancelIndexing(): IndexStatus {
     indexStatusService.markBuildCancelled()
-    hmppsQueueService.purgeQueue(PurgeQueueRequest(indexQueue.queueName, indexQueue.sqsClient, indexQueue.queueUrl))
+    runBlocking {
+      hmppsQueueService.purgeQueue(PurgeQueueRequest(indexQueue.queueName, indexQueue.sqsClient, indexQueue.queueUrl))
```
petergphillips, line 172: `purgeQueue` is a suspending function so need to do a `runBlocking` here

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/config/AuthenticationHolderTest.kt
```diff
@@ -22,6 +23,9 @@ class AuthenticationHolderTest {
     assertThat(authenticationHolder.currentClientId()).isEqualTo("clientID")
   }
 
+  @AfterEach
+  fun afterEach() = SecurityContextHolder.clearContext()
```
petergphillips, line 27: this caused me quite a bit of grief as it had a knock on effect on other tests

andymarke (approved):
lgtm

PR #30, opened 2023-06-21
SDIT-405: ✨ Add in index comparison by size

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/integration/resource/CompareIndexResourceApiTest.kt
```diff
@@ -15,61 +19,103 @@ import uk.gov.justice.digital.hmpps.prisonersearchindexer.integration.wiremock.P
 import uk.gov.justice.digital.hmpps.prisonersearchindexer.model.SyncIndex
 
 class CompareIndexResourceApiTest : IntegrationTestBase() {
+  @Nested
+  inner class compareTestsNoData {
```
petergphillips, line 23: not happy with this test as `IntegrationTestBase` does a clear and re-create index for each test which means that I can't combine data sets and then do a number of comparisons on them.  Will re-write in future PR

mikehalmamoj (approved):

PR #36, opened 2023-06-28
SDIT-405: 🔒️ Remove gson from main dependencies

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/services/NomisService.kt
```diff
@@ -32,12 +32,12 @@ class NomisService(
     }
     .retrieve()
     .bodyToMono(PrisonerNumberPage::class.java)
-    .block() ?: PrisonerNumberPage()
+    .block()!!
```
petergphillips, line 35: improvement from previous PR

mikehalmamoj (approved):

PR #37, opened 2023-06-30
SDIT-405: ✨ Add in repository get and delete methods

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearchindexer/repository/PrisonerRepository.kt
```diff
@@ -29,6 +29,15 @@ class PrisonerRepository(
     openSearchRestTemplate.index(IndexQueryBuilder().withObject(prisoner).build(), IndexCoordinates.of(index.indexName))
   }
 
+  fun delete(prisonerNumber: String) {
+    listOf(SyncIndex.GREEN, SyncIndex.BLUE).forEach {
```
petergphillips, line 33: note that delete removes from all indexes irrespective of which one is active

mikehalmamoj (approved):

PR #45, opened 2023-07-07
SDIT-405: ✨ Add in helm lint and fix test reports

petergphillips:


model/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/model/diff/PrisonerDifferenceService.kt
```diff
@@ -14,13 +14,9 @@ enum class DiffCategory {
   IDENTIFIERS, PERSONAL_DETAILS, ALERTS, STATUS, LOCATION, SENTENCE, RESTRICTED_PATIENT, INCENTIVE_LEVEL, PHYSICAL_DETAILS
 }
 
-data class Difference(val property: String, val categoryChanged: DiffCategory, val oldValue: Any?, val newValue: Any?)
```
petergphillips, line 17: not used by common model - only by indexer

.circleci/config.yml
```diff
@@ -15,5 +15,6 @@ workflows:
       - path-filtering/filter:
           name: check-for-changes
           mapping: |
-            (model/.*|hmpps-prisoner-search/.*) workflow-directory "hmpps-prisoner-search"
```
petergphillips, line 18: turns out that this previous approach was nonsense. `path-filtering` will only run the workflow once so the variable was being set and then overwritten.  Need to instead set multiple variables and then handle in the config

.circleci/continue_config.yml
```diff
@@ -7,12 +7,42 @@ parameters:
...
+    default: false
+  build-indexer:
+    type: boolean
+    default: false
+  build-search:
+    type: boolean
+    default: false
 
 jobs:
+  validate-model:
```
petergphillips, line 21: we end up having lots of separate jobs here instead

.circleci/continue_config.yml
```diff
@@ -39,35 +61,43 @@ jobs:
...
   version: 2
-  build-test-and-deploy:
+  build-test-and-deploy-model:
+    when: << pipeline.parameters.build-model >>
+    jobs:
+      - validate-model:
+          filters:
+            tags:
+              ignore: /.*/
+  build-test-and-deploy-indexer:
```
petergphillips, line 86: note that we end up with 3 separate workflows - one for model, one for search and one for indexer

PR #54, opened 2023-07-11
SDIT-405: 🔧 Build everything if circleci or build gradle changes

petergphillips:


.circleci/continue_config.yml
```diff
@@ -74,12 +74,12 @@ jobs:
       - store_artifacts:
           path: << parameters.project >>/build/reports/tests
 
-  noop:
+  no-op:
```
petergphillips, line 77: this was annoying me

.circleci/continue_config.yml
```diff
@@ -102,7 +102,7 @@ workflows:
   build-test-and-deploy-indexer:
     when: << pipeline.parameters.build-indexer >>
     jobs:
-      - validate:
+      - validate-indexer:
```
petergphillips, line 105: nicer if these are unique and then we don't get the `-1` suffix etc

mikehalmamoj (approved):

PR #56, opened 2023-07-11
SDIT-937 Add Global Search

karenmillermoj:
Initial files added - adding tests in next PR


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/GlobalSearchService.kt
```diff
@@ -0,0 +1,205 @@
...
+        }
+      } else {
+        queryBy(globalSearchCriteria, pageable) { nameMatch(it) } onMatch {
+          customEventForFindBySearchCriteria(globalSearchCriteria, it.matches.size)
+          return PageImpl(it.matches, pageable, it.totalHits)
+        }
+      }
+    }
+    return PageImpl(listOf(), pageable, 0L)
+  }
```
mikehalmamoj, line 54: As this is the top level method for this class it might be worth spending some time making it more readable. For example if you did a `with(globalSearchCriteria) {` you'd lose a lot of noise instantly.

karenmillermoj: good idea - it was just a lift and shift from prisoner-offender-search but the next few bits of work I will be looking at tidying things up

PR #58, opened 2023-07-12
SDIT-937 Set up Index Alias and use config from model

karenmillermoj:


model/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/config/OpenSearchIndexConfiguration.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.config
+
+import org.springframework.context.annotation.Configuration
+
+@Configuration
```
petergphillips, line 5: don't think this needs a `Configuration` annotation?

petergphillips (approved):
LGTM

PR #62, opened 2023-07-12
SDIT-937 Add prisoner-search files

karenmillermoj:
Prisoner search files pulled in from prisoner-offender-search (and modified for opensearch)

### DIFF
--- a/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/GsonConfig.kt
+++ b/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/GsonConfig.kt
@@ -1,0 +1,56 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.search.config
+
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.google.gson.JsonPrimitive
+import com.google.gson.JsonSerializationContext
+import com.google.gson.JsonSerializer
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.lang.reflect.Type
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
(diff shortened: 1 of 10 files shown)


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/PhysicalDetailRequest.kt
```diff
@@ -0,0 +1,162 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.search.services.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
```
karenmillermoj, line 4: This is file is part of the next PR (accidentally included it - but will not affect anything)

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/GsonConfig.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.lang.reflect.Type
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+import java.util.Locale
+
+@Configuration
+class GsonConfig() {
```
petergphillips, line 20: I actually removed this file from indexer and switched to object mapper instead.  Jackson is the default with spring boot so made more sense to just use that out of the box

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/PhysicalDetailRequest.kt
```diff
@@ -0,0 +1,162 @@
...
+data class PhysicalDetailRequest(
+  @Schema(
+    description = "List of prison codes to filter results by",
+    example = "['LEI', 'MDI']",
+    required = true,
```
petergphillips, line 9: think this `required` is now deprecated?  Replaced with `RequiredMode`?

petergphillips (approved):
LGTM

PR #67, opened 2023-07-13
SDIT-405: 🐛 Fix indexing

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/NomisService.kt
```diff
@@ -54,5 +54,5 @@ class NomisService(
 
 data class PrisonerNumberPage(
   val content: List<String> = emptyList(),
-  val totalRecords: Long = 0,
+  val totalElements: Long = 0,
```
petergphillips, line 57: the wiremock test uses this class too hence didn't spot that Prison API is returning something else

karenmillermoj (approved):
lgtm

PR #69, opened 2023-07-14
SDIT-405: ✨ Get comparison to return size too

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/resource/CompareIndexResource.kt
```diff
@@ -26,7 +26,7 @@ class CompareIndexResource(private val compareIndexService: CompareIndexService)
         Kubernetes CronJob named `synthethic-monitor-cronjob`
       """,
   )
-  suspend fun compareIndexSizes(): Unit = compareIndexService.doIndexSizeCheck()
+  fun compareIndexSizes() = compareIndexService.doIndexSizeCheck()
```
petergphillips, line 29: `suspend` must have been left over from when this project was reactive

andymarke (approved):
lgtm

PR #70, opened 2023-07-14
SDIT-405: ✨ Maintenance cronjobs

petergphillips:


hmpps-prisoner-search-indexer/helm_deploy/hmpps-prisoner-search-indexer/templates/check-indexing-complete.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "app.fullname" . }}-check-indexing-complete
```
petergphillips, line 4: this was joined with the retry-dlq cronjob into a single job.  Have split it out into two as each are completely independent - when one fails it shouldn't fail the other etc.

andymarke (approved):
lgtm

PR #72, opened 2023-07-14
SDIT-405: 🐛 Don't output null values

petergphillips:


hmpps-prisoner-search-indexer/src/main/resources/application.yml
```diff
@@ -12,6 +12,7 @@ spring:
     date-format: "yyyy-MM-dd HH:mm:ss"
     serialization:
       WRITE_DATES_AS_TIMESTAMPS: false
+    default-property-inclusion: non_null
```
andymarke, line 15: I always forget to add this!

andymarke (approved):
lgtm

PR #75, opened 2023-07-18
SDIT-937 Add more resource tests

karenmillermoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/GlobalSearchResource.kt
```diff
@@ -48,7 +54,48 @@ class GlobalSearchResource(
...
-  */
+  fun findByPrisonNumber(@PathVariable id: String): Prisoner? =
+    prisonerSearchService.findBySearchCriteria(SearchCriteria(id, null, null)).firstOrNull()
+      ?: throw NotFoundException("$id not found")
 }
+
+/*
+ TODO Decide where this should live now that search and index are separate
+ Have left here for now to ensure this is not forgotten
+
```
karenmillermoj, line 65: Just checked probation - it should be in the search (not indexer so will do a separate PR to pull that in)

mikehalmamoj: Another thing on my todo list is to split out the index count from the synthetic monitor. They're completely different things.

PR #86, opened 2023-07-24
SDIT-405: ✨ Output cron check status

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/resource/MaintainIndexResource.kt
```diff
@@ -167,7 +167,7 @@ class MaintainIndexResource(private val maintainIndexService: MaintainIndexServi
       `index-housekeeping-cronjob`
       """,
   )
-  fun checkIfComplete() {
+  fun checkIfComplete() =
```
petergphillips, line 170: the cronjob was previously not returning any output.  Since we've got the index status we might as well return that

brightonsbox (approved):
LGTM

PR #88, opened 2023-07-24
SDIT-405: ♻️  Finish removing either and just throw

petergphillips:


andymarke (approved):
LGTM

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/PopulateIndexListener.kt
```diff
@@ -32,19 +32,22 @@ class PopulateIndexListener(
...
+        else -> {
+          "Unknown request type for message $requestJson"
+            .let {
+              log.error(it)
+              throw IllegalArgumentException(it)
+            }
+        }
       }
+    } catch (e: ResponseStatusException) {
+      log.error("Message {} failed with error {}", indexRequest, e.message)
```
mikehalmamoj, line 49: This feels like the wrong kind of exception for handling messages - it has a constructor with an HTTP status. However I can see how it's convenient for reusing custom exceptions many of which seem to occur for both message / HTTP request processing.

Also, shouldn't this throw something to make the message be retried?

PR #125, opened 2023-09-12
SDIT-1062: ✨ Add in new refresh of prisoners

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerDifferenceService.kt
```diff
@@ -75,20 +78,14 @@ class PrisonerDifferenceService(
     }
   }
 
-  fun handleDifferencesForReport(previousPrisonerSnapshot: Prisoner?, prisoner: Prisoner) {
```
petergphillips, line 78: unused

lbennett-moj (approved):

PR #158, opened 2023-10-16
SDIT-1150: 🐛 Stop dynamic uri references to prevent tag explosion and running out of memory

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/NomisService.kt
```diff
@@ -33,22 +33,22 @@ class NomisService(
     getOffendersIds(page, pageSize).content
 
   fun getNomsNumberForBooking(bookingId: Long): String? = prisonApiWebClient.get()
-    .uri("/api/bookings/$bookingId?basicInfo=true")
+    .uri("/api/bookings/{bookingId}?basicInfo=true", bookingId)
```
petergphillips, line 36: stop using kotlin interpolation and parameterize calls so that micrometer doesn't generate a separate tag for each uri (and therefore doesn't run out of memory)

mikehalmamoj (approved):

PR #160, opened 2023-10-16
SDIT-1114: 🔧 Use latest gradle plugin to sort suppressions

petergphillips:


dps-gradle-spring-boot-suppressions.xml
```diff
@@ -1,826 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
petergphillips, line 1: doesn't need to be checked in as dps gradle plugin will create each time now instead

karenmillermoj (approved):
lgtm

PR #164, opened 2023-10-23
SDIT-1168: ✨ Turn events on for preprod

petergphillips:


hmpps-prisoner-search-indexer/helm_deploy/values-prod.yaml
```diff
@@ -9,6 +9,7 @@ generic-service:
     API_BASE_URL_RESTRICTED_PATIENTS: "https://restricted-patients-api.hmpps.service.justice.gov.uk"
     API_BASE_URL_INCENTIVES: "https://incentives-api.hmpps.service.justice.gov.uk"
     DIFF_HOST: "https://prisoner-search.prison.service.justice.gov.uk"
+    DIFF_PREFIX: "dummy."
```
petergphillips, line 12: this is the default anyway so this is a no-op.  However it does then make it easier to amend the deployment when needed as the variable is already defined.

mikehalmamoj (approved):

PR #165, opened 2023-10-23
SDIT-1173: 🐛 Ensure that opensearch no change event always raised

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -36,6 +36,8 @@ class PrisonerSynchroniserService(
     if (prisonerDifferenceService.prisonerHasChanged(existingPrisoner, prisoner)) {
       indices.map { index -> prisonerRepository.save(prisoner, index) }
       prisonerDifferenceService.handleDifferences(existingPrisoner, ob, prisoner)
+    } else {
+      telemetryClient.trackPrisonerEvent(TelemetryEvents.PRISONER_OPENSEARCH_NO_CHANGE, ob.offenderNo)
```
petergphillips, line 40: we only call `handleDifferences` if there are changes to the hash, so need to report here that there is no change

andymarke (approved):
lgtm

PR #171, opened 2023-10-24
SDIT-1130: 📝 Document interesting queries

petergphillips:


hmpps-prisoner-search-indexer/UsefulQueries.md
```diff
@@ -1,50 +1,49 @@
...
+Most of the interesting information as to what is happening in the indexer is written as `customEvents`:
+```kusto
+customEvents
+| where cloud_RoleName == "hmpps-prisoner-search-indexer"
 | order by timestamp desc
 ```
+There are 3 main events (`name` field):
+* `PRISONER_UPDATED` if a prisoner has changed
+* `PRISONER_OPENSEARCH_NO_CHANGE` if the OpenSearch index hash already contains the changes
+* `PRISONER_DATABASE_NO_CHANGE` if the OpenSearch index is different but the database hash is the same.
```
mikehalmamoj, line 13: I'm finding this hard to follow... Is there a difference between the "Opensearch index hash" and the "database hash"? What does it mean if the "OpenSearch index is different but the database hash is the same"?

mikehalmamoj (approved):

hmpps-prisoner-search-indexer/UsefulQueries.md
petergphillips: fair point - will improve in next PR and try to explain in slack

PR #176, opened 2023-10-26
SDIT-1177: ✨ Delay processing assessment updated messages by 5 seconds

petergphillips:


steverendell (approved):
lgtm

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/OffenderEventQueueService.kt
```diff
@@ -0,0 +1,40 @@
...
+  }
+  private val offenderEventSqsClient by lazy { offenderEventQueue.sqsClient }
+  private val offenderEventQueueUrl by lazy { offenderEventQueue.queueUrl }
+
+  fun republishMessageWithDelay(message: String, eventType: String) {
+    offenderEventSqsClient.sendMessage(
+      SendMessageRequest.builder().queueUrl(offenderEventQueueUrl)
+        // replace the event type in the raw JSON message. If we parsed the message first then we can't guarantee that
+        // we've converted all the fields in the original JSON as the parser ignores unknown fields.
+        .messageBody(message.replace(eventType, "${eventType}_$REPUBLISH_SUFFIX"))
```
mikehalmamoj, line 28: Isn't there also an event type in the message attributes? Should that also have the republish suffix?

petergphillips: I'm passing in the `requestJson` to this method, which will contain both the message and message attributes fields.  `replace` in kotlin replaces all occurrences, not just the first one (as confirmed by the test case)

PR #182, opened 2023-10-30
SDIT-1129: 🐛 Remove health check as unnecessary (only calls auth once for jwts

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/WebClientConfiguration.kt
```diff
@@ -1,27 +0,0 @@
...
-import org.springframework.context.annotation.Bean
-import org.springframework.context.annotation.Configuration
-import org.springframework.http.client.reactive.ReactorClientHttpConnector
-import org.springframework.scheduling.annotation.EnableAsync
-import org.springframework.web.reactive.function.client.WebClient
-import reactor.netty.http.client.HttpClient
-import java.time.Duration
-
-@Configuration
-@EnableAsync
```
petergphillips, line 13: async not used in this repo anyway

hmpps-prisoner-search/src/main/resources/application.yml
```diff
@@ -22,11 +22,6 @@ spring:
         jwt:
           jwk-set-uri: ${api.base.url.oauth}/.well-known/jwks.json
```
petergphillips, line 23: jwts is cached forever by this app

andymarke (approved):
lgtm

PR #196, opened 2023-11-22
SDIT-1262: 🐛 Try wiring elasticsearchOperations instead

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/health/OpenSearchHealthIndicator.kt
```diff
@@ -1,15 +1,14 @@
...
 import org.springframework.boot.actuate.health.Health
 import org.springframework.boot.actuate.health.Status
+import org.springframework.data.elasticsearch.core.ElasticsearchOperations
 import org.springframework.data.elasticsearch.core.cluster.ClusterHealth
 import org.springframework.stereotype.Component
 
-@Suppress("SpringJavaInjectionPointsAutowiringInspection")
 @Component
-class OpenSearchHealthIndicator(private val template: OpenSearchRestTemplate) : AbstractHealthIndicator() {
+class OpenSearchHealthIndicator(private val template: ElasticsearchOperations) : AbstractHealthIndicator() {
```
petergphillips, line 11: it couldn't find a `OpenSearchRestTemplate`.  Turns out that only `elasticsearchoperations` is actually returned and wired up - https://github.com/opensearch-project/spring-data-opensearch/blob/main/spring-data-opensearch/src/main/java/org/opensearch/data/client/orhlc/AbstractOpenSearchConfiguration.java#L37

mikehalmamoj (approved):

PR #198, opened 2023-11-24
SDIT-1262: 🐛 Change index state to absent until new index created

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/MaintainIndexService.kt
```diff
@@ -44,9 +44,10 @@ class MaintainIndexService(
   }
 
   private fun doPrepareIndexForRebuild(indexStatus: IndexStatus): IndexStatus {
-    indexStatusService.markBuildInProgress()
+    indexStatusService.markBuildAbsent()
```
petergphillips, line 47: When the build is in progress then queue listeners can process messages and write prisoners to the index.
We want to hold off doing that until the index has been created properly

mikehalmamoj: Not sure if we need a test for this behaviour or how hard that would be?

mikehalmamoj (approved):

PR #214, opened 2024-01-16
SDIT-1361 Fix test but removing unnecessary wait

andymarke:


hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/HmppsDomainEventsEmitterIntTest.kt
```diff
@@ -247,7 +247,6 @@ class HmppsDomainEventsEmitterIntTest : IntegrationTestBase() {
     val message = "/messages/offenderDetailsChanged.json".readResourceAsText().replace("A7089FD", "A1239DD")
 
     offenderQueueSqsClient.sendMessage(message)
-    await untilCallTo { getNumberOfMessagesCurrentlyOnEventQueue() } matches { it == 0 }
```
andymarke, line 250: if this is slow, and polls bad, the 1 second expected delay won't happen below since this wait is eating into the 1 second delay

andymarke: Also it adds nothing to the test

PR #224, opened 2024-02-13
SDIT-1490 Design a basic attribute search request

mikehalmamoj:


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequestTest.kt
```diff
@@ -0,0 +1,616 @@
...
+                "textMatchers": [
+                  {
+                    "attribute": "firstName",
+                    "condition": "IS",
+                    "searchTerm": "John"
+                  }
+                ]
+              }
+            ]
+          }""".trimIndent())
```
mikehalmamoj, line 30: Best to read the test name and the JSON to perform the search. The assertions afterwards just prove that the JSON is parsed into our query objects correctly.

With a single value the `joinType` is kind of irrelevant but didn't want to default it to AND. Maybe we should?

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
```diff
@@ -0,0 +1,60 @@
...
+  val bodyPart: String? = null,
+  val comment: String? = null,
+)
+
+enum class JoinType {
+  AND,
+  OR,
+}
+
+enum class TextCondition {
```
petergphillips, line 57: don't think this is just a text condition - e.g. hair colour **IS** blonde

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
```diff
@@ -0,0 +1,60 @@
...
+  val maxValue: Int? = null,
+  val maxInclusive: Boolean = true,
+)
+
+data class DateMatcher(
+  val attribute: String,
+  val minValue: LocalDate? = null,
+  val minInclusive: Boolean = true,
+  val maxValue: LocalDate? = null,
+  val maxInclusive: Boolean = true,
```
petergphillips, line 43: looking at the UI don't think we need the min or max inclusive options

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
mikehalmamoj: I decided to treat the enums as just text - they're just Strings on the Prisoner record. The fact the text is predetermined in a drop down is neither here nor there, just increases the likeihood of exact matches.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+data class BooleanMatcher(
+  val attribute: String,
+  val condition: Boolean,
+)
+
+data class IntegerMatcher(
+  val attribute: String,
+  val minValue: Int? = null,
+  val minInclusive: Boolean = true,
```
petergphillips, line 33: looking at the UI - the new figma has a few examples, but don't think we need the min or max inclusive options

petergphillips: although it is difficult to tell - the latest one has an **IS** condition with cm that doesn't make any sense

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequestTest.kt
```diff
@@ -0,0 +1,649 @@
...
+    )
+  }
+
+  @Test
+  fun `firstName is John AND lastName is Smith`() {
+    val request = objectMapper.readValue<AttributeSearchRequest>(
+      """{
+            "matchers": [
+              {
+                "joinType": "AND",
```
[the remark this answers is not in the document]
mikehalmamoj: I didn't want to just follow the format they have currently come up with because it's very much tied to their implementation. This design is flexible and can accomodate changes.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
mikehalmamoj: I couldn't see any examples for integer based fields. Do you think it should be LT, LTE, EQ, GT, GTE?

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
mikehalmamoj: Always inclusive then?

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/dto/AttributeSearchRequest.kt
petergphillips: difficult to tell which way they'll go. Looking at their front end for dates it has both dates on the same row, so doesn't allow inclusive (or is it exclusive) options

PR #231, opened 2024-02-16
SDIT-1491: ✨ Add caching of reference data

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/ReferenceDataService.kt
```diff
@@ -18,29 +19,29 @@ class ReferenceDataService(
   @Value("\${search.detailed.max-results}") private val maxSearchResults: Int = 200,
   @Value("\${search.detailed.timeout-seconds}") private val searchTimeoutSeconds: Long = 10L,
 ) {
+  // we have implemented a cache loader in CacheConfig so that will call the findReferenceData method below
+  @Cacheable(cacheNames = ["referenceData"], key = "#attribute")
+  fun findReferenceDataCached(attribute: ReferenceDataAttribute) = ReferenceDataResponse()
```
petergphillips, line 24: this is odd - but because we are using a loading cache the implementation will never be called

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/ReferenceDataService.kt
```diff
@@ -18,29 +19,29 @@ class ReferenceDataService(
...
 
-    return try {
-      val searchResponse = elasticsearchClient.search(searchRequest)
-      val aggregation: MultiBucketsAggregation = searchResponse.aggregations.get(attribute.name)
-      ReferenceDataResponse(
-        aggregation.buckets.map {
-          ReferenceData(it.keyAsString)
-        }.sortedBy { it.key },
-      )
-    } catch (e: Throwable) {
```
petergphillips, line 33: this looked wrong - don't like the idea of catching all throwables and then just returning no results back to the client

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/ReferenceDataResourceTest.kt
```diff
@@ -35,9 +42,15 @@ import uk.gov.justice.digital.hmpps.prisonersearch.search.services.ReferenceData
 import uk.gov.justice.digital.hmpps.prisonersearch.search.services.ReferenceDataAttribute.status
 import uk.gov.justice.digital.hmpps.prisonersearch.search.services.ReferenceDataAttribute.youthOffender
 import uk.gov.justice.digital.hmpps.prisonersearch.search.services.ReferenceDataResponse
+import uk.gov.justice.digital.hmpps.prisonersearch.search.services.SearchClient
 import java.util.stream.Stream
 
 class ReferenceDataResourceTest : AbstractSearchDataIntegrationTest() {
+  @SpyBean
```
mikehalmamoj, line 49: Maybe move this into the super class so Spring doesn't have to restart the app due to a changed Context?

mikehalmamoj (approved):

PR #239, opened 2024-02-20
SDIT-1490 Use queries and subqueries to describe joined type matchers

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/api/Query.kt
```diff
@@ -0,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.search.services.attributesearch.api
+
+import uk.gov.justice.digital.hmpps.prisonersearch.search.services.attributesearch.AttributeSearchException
+
+data class Query(
+  val joinType: JoinType,
+  val matchers: List<TypeMatcher<*>>? = null,
+  val subQueries: List<Query>? = null,
```
mikehalmamoj, line 8: Instead of Matchers / children we now call each list of joins as Queries / subQueries

petergphillips (approved):
LGTM

PR #243, opened 2024-02-21
SDIT-1509 await publish so that any errors are thrown

andymarke:
By doing the join and therefore waiting for the publish() result, any errors are thrown which means they will be retired and eventually move to DLQ

### DIFF
--- a/hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventPublisherListenerTest.kt
+++ b/hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventPublisherListenerTest.kt
@@ -1,0 +1,101 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.indexer.listeners
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import kotlinx.coroutines.test.runTest
+import org.awaitility.kotlin.await
+import org.awaitility.kotlin.matches
+import org.awaitility.kotlin.untilCallTo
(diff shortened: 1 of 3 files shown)


hmpps-prisoner-search-indexer/src/test/resources/application-test.yml
```diff
@@ -62,7 +62,7 @@ hmpps.sqs:
     publish:
       queueName: ${random.uuid}
       dlqName: ${random.uuid}
-      dlqMaxReceiveCount: 1
+      dlqMaxReceiveCount: 2
   topics:
```
andymarke, line 66: make sure there are retries when testing

hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventPublisherListenerTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import uk.gov.justice.hmpps.sqs.MissingQueueException
+import uk.gov.justice.hmpps.sqs.PurgeQueueRequest
+import uk.gov.justice.hmpps.sqs.countAllMessagesOnQueue
+import java.util.concurrent.CompletableFuture
+
+class DomainEventPublisherListenerTest : IntegrationTestBase() {
+
+  @SpyBean
+  @Qualifier("hmppseventtopic-sns-client")
+  private lateinit var hmppsEventTopicSnsClient: SnsAsyncClient
```
mikehalmamoj, line 32: Better in the test base class so Spring doesn't start a new context?

mikehalmamoj (approved):

PR #251, opened 2024-02-22
SDIT-1493 Add last prison ID to the prisoner record

mikehalmamoj:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
```diff
@@ -122,6 +122,11 @@ open class Prisoner : Diffable<Prisoner> {
   @DiffableProperty(DiffCategory.LOCATION)
   var prisonId: String? = null
 
+  @Field(type = FieldType.Keyword)
+  @Schema(description = "The last location of the prisoner", example = "MDI")
```
petergphillips, line 126: would be good to expand on what this means - if they got transferred from one prison to another one will this value be set?

petergphillips (approved):
LGTM

PR #254, opened 2024-02-26
SDIT-1491: ✨ Add in prison api to search so can call it for reference data

petergphillips:


hmpps-prisoner-search/src/main/resources/application.yml
```diff
@@ -20,7 +20,19 @@ spring:
...
+        registration:
+          prison-api:
+            provider: hmpps-auth
+            client-id: ${prison-api.client.id}
+            client-secret: ${prison-api.client.secret}
+            authorization-grant-type: client_credentials
+            scope: read
+        provider:
+          hmpps-auth:
+            token-uri: ${api.base.url.oauth}/oauth/token
```
karenmillermoj, line 35: shouldn't token-uri be the new hmpps-auth.base-uri etc

petergphillips: 🤦 - I thought I'd fixed all the references

karenmillermoj (approved):
lgtm

PR #258, opened 2024-02-27
SDIT-1491: ✨ Return label / value pairs for reference data

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/ReferenceDataService.kt
```diff
@@ -29,11 +27,12 @@ class ReferenceDataService(
...
     val searchResponse = elasticsearchClient.search(searchRequest)
     val aggregation: MultiBucketsAggregation = searchResponse.aggregations.get(attribute.name)
-    return ReferenceDataResponse(
-      aggregation.buckets.map {
-        ReferenceData(it.keyAsString)
-      }.sortedBy { it.key },
-    )
+    return aggregation.buckets.map {
+      val key = it.keyAsString
+      ReferenceData(value = key, label = attribute.map?.get(key) ?: key)
```
petergphillips, line 32: default the label to the key if we haven't mapped it

mikehalmamoj (approved):

PR #259, opened 2024-02-27
SDIT-1336 Go back to old api URL properties - it's established and it works fine

mikehalmamoj:
And it means you can do things like [this](https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/main/helm_deploy/values-prod.yaml#L12) and [this](https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/main/src/test/resources/application-test.yml#L8). Which are easier to digest.


petergphillips (approved):
LGTM

PR #270, opened 2024-03-06
SDIT-1561 Fix bug with times on zero seconds

mikehalmamoj:
DateTimeMatcher tests intermittently failed when the time had zero seconds because it was being truncated as per the LocalDateTime.toString() Javadocs:
```
 The output will be one of the following ISO-8601 formats:

    uuuu-MM-dd'T'HH:mm
    uuuu-MM-dd'T'HH:mm:ss
    uuuu-MM-dd'T'HH:mm:ss.SSS
    uuuu-MM-dd'T'HH:mm:ss.SSSSSS
    uuuu-MM-dd'T'HH:mm:ss.SSSSSSSSS

The format used will be the shortest that outputs the full value of the time where the omitted parts are implied to be zero.
```


petergphillips (approved):
LGTM

PR #271, opened 2024-03-06
SDIT-1561 Try out parallel search tests

mikehalmamoj:
Split out attribute search tests from the other tests and run them in parallel.
image

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -52,7 +52,31 @@ jobs:
       - store_artifacts:
           path: common/build/reports/tests
 
-  validate-search:
+  validate-search-check:
+    parameters:
+      project:
+        type: string
+    executor:
+      name: hmpps/java
+      tag: "21.0"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "<< parameters.project >>/build.gradle.kts" }}
+            - gradle-
+      - run:
+          command: ./gradlew :<< parameters.project >>:check -x test
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: gradle-{{ checksum "<< parameters.project >>/build.gradle.kts" }}
+      - store_test_results:
+          path: << parameters.project >>/build/test-results
+      - store_artifacts:
+          path: << parameters.project >>/build/reports/tests
+
+  validate-search-test-attribute-search:
     parameters:
       project:
         type: string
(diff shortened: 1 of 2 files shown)


hmpps-prisoner-search/build.gradle.kts
```diff
@@ -77,3 +77,35 @@ tasks {
...
+
+  useJUnitPlatform()
+
+  // Override test class exclude patterns if necessary
+  excludes.clear()
+  exclude("**/AttributeSearchIntegrationTest.*")
+
+  // Force the task to always run
+  outputs.upToDateWhen { false }
+}
```
mikehalmamoj, line 111: 2 new Gradle tasjks - one to `include("**/AttributeSearchIntegrationTest.*")` and one to `exclude("**/AttributeSearchIntegrationTest.*")`.

This means I don't have to mess with the `test` task - it still runs all the tests locally if that's what you want to do.

.circleci/continue_config.yml
```diff
@@ -52,7 +52,31 @@ jobs:
...
+      name: hmpps/java
+      tag: "21.0"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "<< parameters.project >>/build.gradle.kts" }}
+            - gradle-
+      - run:
+          command: ./gradlew :<< parameters.project >>:check -x test
```
mikehalmamoj, line 69: When running the `check` task ignore the tests - we'll run them separately.

.circleci/continue_config.yml
```diff
@@ -77,7 +101,42 @@ jobs:
             - gradle-{{ checksum "<< parameters.project >>/build.gradle.kts" }}
             - gradle-
       - run:
-          command: ./gradlew :<< parameters.project >>:check
+          command: ./gradlew :<< parameters.project >>:attributeSearchIntegrationTest
```
mikehalmamoj, line 104: A separate job that runs only the attribure search tests.

.circleci/continue_config.yml
petergphillips: I find this confusing - there is a `validate-search-check` task that doesn't actually run all the checks?  What does it do then?

hmpps-prisoner-search/build.gradle.kts
```diff
@@ -77,3 +77,35 @@ tasks {
...
+  // Inherit configurations from the existing 'test' task
+  val testTask = tasks.named("test").get() as Test
+  testClassesDirs = testTask.testClassesDirs
+  classpath = testTask.classpath
+
+  useJUnitPlatform()
+
+  // Override test class include patterns if necessary
+  includes.clear()
+  include("**/AttributeSearchIntegrationTest.*")
```
petergphillips, line 91: why the `.*` at the end?  I think there is only one attribute search class?

PR #275, opened 2024-03-11
SDIT-1561 Support paging and sorting

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeSearchService.kt
```diff
@@ -45,13 +48,36 @@ class AttributeSearchService(
...
+    val sortBuilders = sort.map {
+      FieldSortBuilder(getSearchableProperty(it)).order(SortOrder.fromString(it.direction.name))
+    }.toList()
+    return SearchSourceBuilder().apply {
+      query(queryBuilder)
+      size(pageSize)
+      from(offset.toInt())
+      sortBuilders.takeIf { it.isNotEmpty() }
+        ?.forEach { sort(it) }
+        ?: sort("prisonerNumber")
```
mikehalmamoj, line 68: As we implement default paging (pageSize=10) we also need a default sort order.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeSearchService.kt
```diff
@@ -45,13 +48,36 @@ class AttributeSearchService(
...
+  }
+
+  private fun getSearchableProperty(it: Sort.Order): String? {
+    val type = attributes[it.property] ?: throw AttributeSearchException("Sort attribute '${it.property}' not found")
+    return when {
+      it.property == "prisonerNumber" -> "prisonerNumber"
+      type.simpleName == "String" -> "${it.property}.keyword"
+      else -> it.property
+    }
+  }
```
mikehalmamoj, line 79: Strings need sorting by the keyword - except for prisonerNumber (which I guess is because it's the `@Id` column).

PR #279, opened 2024-03-14
SDIT-1561 String searches fine tuning

mikehalmamoj:
All string searches should be case-insensitive. String contains searches should allow single character (?) and multi-character (*) wildcards.


andymarke (approved):
lgtm

PR #282, opened 2024-03-18
SDIT-1561 Handle lists of nested objects

mikehalmamoj:
Found a bug while testing searching on physical details - if you search for prisoners with both `tattoos.bodyPart=Arm` and `tattoos.comment=dragon` then you're asking for prisoners with a tattoo of a dragon on their arm. However you also receive anyone with 2 tattoos, one on their arm and the other of a dragon. This is because the BodyPartDetails class does not have the `@Field(type = FieldType.Nested)` annotation, meaning that [OpenSearch flattens the fields](https://opensearch.org/docs/latest/field-types/supported-field-types/nested/#flattened-form) and loses any connection between `tattoo.bodyPart` and `tattoo.comment`.

First I've updated the tests for `alerts` (which is already nested) to highlight the bug - and fixed that bug in the query builders by using a NestedQuery.

Next job will be to add the nested annotation for any lists of `BodyPart`s and then rebuilding the index in each environment.

### DIFF
--- a/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
+++ b/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
@@ -16,6 +16,7 @@ typealias Attributes = Map<String, Attribute>
 class Attribute(
   val type: KClass<*>,
   val openSearchName: String,
+  val nested: Boolean = false,
 )
 
 @Configuration
@@ -30,20 +31,21 @@ internal fun getAttributes(kClass: KClass<*>): Attributes =
 private fun findAttributes(
   prop: KProperty1<*, *>,
   prefix: String = "",
+  nested: Boolean = false,
 ): List<Pair<String, Attribute>> =
   when (getPropertyType(prop)) {
     PropertyType.SIMPLE -> {
       getPropertyClass(prop).let { propClass ->
-        listOf("${prefix}${prop.name}" to Attribute(propClass, "${prefix}${getOpenSearchName(prop, propClass)}"))
+        listOf("${prefix}${prop.name}" to Attribute(propClass, "${prefix}${getOpenSearchName(prop, propClass)}", nested))
       }
     }
     PropertyType.LIST -> {
       getGenericTypeClass(prop).memberProperties
-        .flatMap { childProp -> findAttributes(childProp, "${prefix}${prop.name}.") }
+        .flatMap { childProp -> findAttributes(childProp, "${prefix}${prop.name}.", nested || prop.hasFieldAnnotation("Nested")) }
     }
     PropertyType.COMPLEX -> {
       getPropertyClass(prop).memberProperties
-        .flatMap { childProp -> findAttributes(childProp, "${prefix}${prop.name}.") }
+        .flatMap { childProp -> findAttributes(childProp, "${prefix}${prop.name}.", nested || prop.hasFieldAnnotation("Nested")) }
     }
   }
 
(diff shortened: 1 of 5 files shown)


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -16,6 +16,7 @@ typealias Attributes = Map<String, Attribute>
 class Attribute(
   val type: KClass<*>,
   val openSearchName: String,
+  val nested: Boolean = false,
 )
```
mikehalmamoj, line 20: We now keep track of which attributes are part of a nested object.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/api/Query.kt
```diff
@@ -40,6 +39,32 @@ data class Query(
...
+        must(
+          QueryBuilders.nestedQuery(
+            prefix,
+            matchers.fold(QueryBuilders.boolQuery()) { boolQueryBuilder, typeMatcher ->
+              boolQueryBuilder.must(typeMatcher.buildQuery(attributes))
+            },
+            ScoreMode.None,
+          ),
+        )
+      }
```
mikehalmamoj, line 62: Any matchers for the same nested objects are grouped together into a NestedQuery which tells OpenSearch that all the matchers must match within the same object.

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PhysicalDetailResourceTest.kt
```diff
@@ -623,6 +623,16 @@ class PhysicalDetailResourceTest : AbstractSearchIntegrationTest() {
...
 
+    // TODO I think this is wrong - neither prisoner has a single tattoo with both Knee and rose
+    @Test
+    fun `searching by tattoos with a body part and a comment on different tattoos`(): Unit = physicalDetailSearch(
+      detailRequest = PhysicalDetailRequest(
+        tattoos = listOf(BodyPart(bodyPart = "Knee", comment = "rose")),
+        prisonIds = listOf("AGI", "LEI"),
+      ),
+      expectedPrisoners = listOf("G7089EZ", "G7090AD"),
+    )
```
mikehalmamoj, line 634: I was wondering why Physical Details Search didn't have the same problem - but it does. So I added a test to highlight the issue that will need updating when we make tattoos a nested object.

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeResolverTest.kt
```diff
@@ -73,6 +75,7 @@ class AttributeResolverTest {
   fun `should resolve attributes`(attributeName: String, attribute: Attribute) {
     assertThat(attributes[attributeName]?.type).isEqualTo(attribute.type)
     assertThat(attributes[attributeName]?.openSearchName).isEqualTo(attribute.openSearchName)
+    assertThat(attributes[attributeName]?.nested).isEqualTo(attribute.nested)
```
mikehalmamoj, line 78: Tests that we can correctly derive whether an attribute is part of a nested object or not.

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeSearchIntegrationTest.kt
```diff
@@ -426,19 +437,19 @@ class AttributeSearchIntegrationTest : AbstractSearchIntegrationTest() {
...
         query {
           stringMatcher("alerts.alertType" IS "AT1")
-          stringMatcher("alerts.alertCode" IS "AC1")
+          stringMatcher("alerts.alertCode" IS "AC2")
         }
       }
 
-      webTestClient.attributeSearch(request).expectPrisoners("P1")
+      webTestClient.attributeSearch(request).expectPrisoners("P2")
     }
```
mikehalmamoj, line 453: Adjusted the test data so that P1 has 1 alert with type of AT1 and another alert with code AC2 - while P2 now has an alert with both type=AT1 and code=AC2.

petergphillips (approved):
LGTM

PR #285, opened 2024-03-19
SDIT-1561 Remove otherMarks from the Prisoner record

mikehalmamoj:
In NOMIS there appears to be no quantifiable difference between `marks` and `otherMarks` - but leaving them separate means they have to be searched for individually. Instead we will delete `otherMarks` and consolidate them into `marks`. It should be safe to delete `otherMarks` because a) it's nullable and b) no clients are referencing the field anyway (except in test JSON).

We've also noticed that there are a lot of comments in `marks` and `otherMarks` which include either "tattoo" or "scar" but both NOMIS and prisoner search have dedicated storage for `tattoos` and `scars`. This makes it hard to search for a tattoo or a scar - you'd need to check marks too. To make the search easier we've decided to copy any marks with comments including "tattoo" or "scar" into the relevant prisoner search field. Note that they will still be included in the marks field so we don't lose any info.

### DIFF
--- a/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/translator.kt
+++ b/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/translator.kt
@@ -67,8 +67,11 @@ fun Prisoner.translate(existingPrisoner: Prisoner? = null, ob: OffenderBooking, 
       when (pm.type) {
         "Tattoo" -> this.tattoos = this.tattoos?.plus(bodyPart) ?: listOf(bodyPart)
         "Scar" -> this.scars = this.scars?.plus(bodyPart) ?: listOf(bodyPart)
-        "Mark" -> this.marks = this.marks?.plus(bodyPart) ?: listOf(bodyPart)
-        "Other" -> this.otherMarks = this.otherMarks?.plus(bodyPart) ?: listOf(bodyPart)
+        "Mark", "Other" -> {
+          this.marks = this.marks?.plus(bodyPart) ?: listOf(bodyPart)
+          this.tattoos = this.tattoos.addIfContains(bodyPart, "tattoo")
+          this.scars = this.scars.addIfContains(bodyPart, "scar")
+        }
       }
     }
   }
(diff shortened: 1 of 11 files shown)


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/translator.kt
```diff
@@ -67,8 +67,11 @@ fun Prisoner.translate(existingPrisoner: Prisoner? = null, ob: OffenderBooking,
       when (pm.type) {
         "Tattoo" -> this.tattoos = this.tattoos?.plus(bodyPart) ?: listOf(bodyPart)
         "Scar" -> this.scars = this.scars?.plus(bodyPart) ?: listOf(bodyPart)
-        "Mark" -> this.marks = this.marks?.plus(bodyPart) ?: listOf(bodyPart)
-        "Other" -> this.otherMarks = this.otherMarks?.plus(bodyPart) ?: listOf(bodyPart)
+        "Mark", "Other" -> {
+          this.marks = this.marks?.plus(bodyPart) ?: listOf(bodyPart)
+          this.tattoos = this.tattoos.addIfContains(bodyPart, "tattoo")
```
petergphillips, line 72: why no `?` on `this.tattoos`?

petergphillips: ah just spotted that `addIfContains` is a function defined below

mikehalmamoj: Because the extension function is on `List<BodyPartDetail>?`

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/translator.kt
```diff
@@ -148,3 +151,12 @@ private fun IncentiveLevel?.toCurrentIncentive(): CurrentIncentive? = this?.let
     dateTime = it.iepTime.withNano(0),
   )
 }
+
+private fun List<BodyPartDetail>?.addIfContains(bodyPart: BodyPartDetail, keyword: String): List<BodyPartDetail>? =
```
petergphillips, line 155: how about
```suggestion
private fun List<BodyPartDetail>?.addIfCommentContains(bodyPart: BodyPartDetail, keyword: String): List<BodyPartDetail>? =
```
I struggled with the function initially - thought it was a built in and didn't realise it was one of ours

petergphillips (approved):
LGTM

PR #289, opened 2024-03-21
SDIT-1561 Add support for fuzzy queries

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/AttributeSearchResource.kt
```diff
@@ -22,7 +22,15 @@ import uk.gov.justice.digital.hmpps.prisonersearch.search.resource.advice.ErrorR
...
-typealias AttributeTypes = Map<String, String>
+@Schema(description = "An attribute that can be searched for in a query")
+data class Attribute(
+  @Schema(description = "The name of the attribute to be used when searching", example = "firstName")
+  val name: String,
+  @Schema(description = "The type of the attribute (used to determine which matcher to use)", example = "String")
+  val type: String,
+  @Schema(description = "Whether the attribute search will be fuzzy. Generally applicable to String attributes containing free text such as names.", example = "true")
+  val fuzzySearch: Boolean,
+)
```
mikehalmamoj, line 33: The get attributes endpoints now tells you if a field supports fuzzy searches.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -13,10 +13,28 @@ import kotlin.reflect.jvm.javaField
...
+  "locationDescription",
+  "dischargeHospitalDescription",
+  "dischargeDetails",
+  "tattoos.comment",
+  "scars.comment",
+  "marks.comment",
+  "aliases.firstName",
+  "aliases.middleNames",
+  "aliases.lastName",
+)
```
mikehalmamoj, line 31: A fuzzy search is only available on attributes where it makes sense - generally free text entered by humans but also some of the longer descriptions.

It doesn't make sense to allow a fuzzy search on any static data such as reference codes or descriptions otherwise the search will turn up many false positives.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/api/StringMatcher.kt
```diff
@@ -25,19 +29,33 @@ data class StringMatcher(
...
+              .should(query)
+              .should(QueryBuilders.fuzzyQuery(attr.openSearchName, searchTerm))
+          }
+          attr.isFuzzy && condition == CONTAINS -> {
+            QueryBuilders.boolQuery()
+              .should(query)
+              .should(QueryBuilders.matchQuery(attribute, searchTerm).fuzziness(Fuzziness.AUTO))
+          }
+          else -> query
+        }
```
mikehalmamoj, line 51: For fuzzy attributes we don't want to lose the standard behaviour, just add a fuzzy search to it.

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeSearchIntegrationTest.kt
```diff
@@ -54,7 +54,7 @@ class AttributeSearchIntegrationTest : AbstractSearchIntegrationTest() {
     ),
     PrisonerBuilder(
       prisonerNumber = "P2",
-      firstName = "John2",
+      firstName = "James2",
```
mikehalmamoj, line 57: Had to make the names more different as they were being picked up by the fuzzy searches.

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AttributeSearchIntegrationTest.kt
```diff
@@ -267,7 +267,7 @@ class AttributeSearchIntegrationTest : AbstractSearchIntegrationTest() {
         }
       }
 
-      webTestClient.attributeSearch(request).expectPrisoners("P1", "P2", "P3")
+      webTestClient.attributeSearch(request).expectPrisoners("P1")
```
mikehalmamoj, line 270: A few of the tests based on names have different expectations due to the name changes.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/AttributeSearchResource.kt
petergphillips: would be good to explain what _fuzzy_ means - maybe at a higher level?

petergphillips: just spotted you've done that below 😂

petergphillips (approved):
LGTM

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/api/StringMatcher.kt
andymarke: @mikehalmamoj does that mean for instance first name will be fuzzy in the this context if you like it or not? So if I search for "Marke" i will get "Marc" back as well and I can't control that behaviour?

mikehalmamoj: Yes that's correct. When we spoke to the product guys the other day they were adamant that it's better to receive too much data rather than miss something.

However it does feel like we should allow it to be turned off for other use cases - just don't have any yet. Do you think I should add a query parm to turn it on and off? What should the default be?

PR #292, opened 2024-03-27
SDIT-1561 Don't perform fuzzy query for wildcards

mikehalmamoj:
Otherwise the fuzzy query will find partial matches on the search term but we want a full match


andymarke (approved):
lgtm

PR #297, opened 2024-04-04
SDIT-1583 Add a client to get all alert types and codes from prison-api

mikehalmamoj:


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonApiServiceIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+  @Nested
+  inner class GetAlerts {
+    @BeforeEach
+    fun setUp() {
+      prisonApi.stubAlertTypes(alertTypes)
+      prisonApi.stubAlertCodes(alertCodes)
+    }
+
+    @Test
+    fun `should get alerts`() {
```
mikehalmamoj, line 21: I'm not sure how far to go with negative tests for this as the data is already there and in a decent shape - though I probably should put something in - but for now I'm all happy pathed up.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonApiService.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import org.springframework.web.reactive.function.client.bodyToMono
+
+@Component
+class PrisonApiService(val prisonApiWebClient: WebClient) {
+
+  fun getAllAlerts(): List<AlertType> {
+    return runBlocking {
+      val alertTypes = async { getDomainReferenceCodes("ALERT") }
```
petergphillips, line 15: I should have mentioned that there is an endpoint in Prison API to return the nested data - https://prison-api-dev.prison.service.justice.gov.uk/swagger-ui/index.html#/reference-domains/getAlertTypes rather than having to get the reference domains separately

petergphillips (approved):
LGTM

PR #299, opened 2024-04-05
SDIT-1583 Use the existing alert types endpoint in prison-api

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonApiService.kt
```diff
@@ -1,38 +1,30 @@
...
+        },
+      )
     }
-  }
 
-  private suspend fun getDomainReferenceCodes(domain: String) = prisonApiWebClient.get()
-    .uri("/api/reference-domains/domains/{domain}/codes", domain)
+  private fun getAlertTypes() = prisonApiWebClient.get()
+    .uri("/api/reference-domains/alertTypes")
+    .header("Page-Limit", "1000")
```
mikehalmamoj, line 24: We have 20 different alert types in prod but in an attack of over-cautiousness I went big.

petergphillips (approved):
LGTM

PR #303, opened 2024-04-08
SDIT-1583 Add cache to alerts reference data endpoint

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/CacheConfig.kt
```diff
@@ -22,10 +22,23 @@ class CacheConfig(private val referenceDataService: ReferenceDataService) : Cach
       .refreshAfterWrite(1, TimeUnit.HOURS)
       // throw away any results after 3 hours so that we don't serve up data that is too stale
       .expireAfterWrite(3, TimeUnit.HOURS)
-      .maximumSize(ReferenceDataAttribute.entries.size + 1L)
```
mikehalmamoj, line 25: This isn't used ([only required for LRU](https://github.com/ben-manes/caffeine/wiki/Eviction#size-based)) so removed for clarity.

petergphillips (approved):
LGTM

PR #314, opened 2024-04-17
SDIT-1644 500 errors returned from prisoner search

steverendell:
Trap the 502 error and retry

### DIFF
--- a/hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/SearchClientTest.kt
+++ b/hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/SearchClientTest.kt
@@ -1,0 +1,91 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.search.services
+
+import org.apache.http.ProtocolVersion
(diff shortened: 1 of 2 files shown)


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/SearchClientTest.kt
```diff
@@ -0,0 +1,91 @@
...
+  private val protocolVersion = ProtocolVersion("HTTP", 1, 1)
+  private val mockResponse = mock<Response>().apply {
+    whenever(this.statusLine).thenReturn(BasicStatusLine(protocolVersion, 502, "Bad Gateway"))
+    whenever(this.requestLine).thenReturn(BasicRequestLine("method", "url", protocolVersion))
+  }
+  private val openSearchStatusException = OpenSearchStatusException(
+    "unable to proxy request",
+    RestStatus.BAD_GATEWAY,
+    ResponseException(mockResponse),
+  )
```
steverendell, line 40: Class Response has a private constructor so a mock seems the only way forward

mikehalmamoj (approved):

PR #319, opened 2024-04-23
SDIT-1679 Don't produce a diff if old endpoint returns null ADA

mikehalmamoj:
Because sometimes it does, sometimes it doesn't.


petergphillips (approved):
LGTM

PR #323, opened 2024-04-25
SDIT-1681: ✨ Bump to latest versions and fail on trivy error

petergphillips:


.github/workflows/security_trivy.yml
```diff
@@ -30,7 +30,8 @@ jobs:
         skip-files: /app/agent.jar
         format: 'sarif'
         output: 'trivy-results.sarif'
-    - uses: github/codeql-action/upload-sarif@v2
+        exit-code: 1
```
petergphillips, line 33: default appears to be `0` (success), but we would prefer to fail the job

PR #324, opened 2024-04-26
SDIT-1681: 🐛 Only output severity defects to sarif format

petergphillips:


.github/workflows/security_trivy.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         format: 'sarif'
         output: 'trivy-results.sarif'
         exit-code: 1
+        limit-severities-for-sarif: true
```
petergphillips, line 34: starting to get complicated.  We're only interested in high and critical issues, but by default the sarif format retrieves all the issues.  This is then compounded by the `exit-code`, that looks at whether any issues are in the output format and doesn't take into account that it is a sarif format.  Would be nice to see all the issues in the security tab, but that doesn't look like it is possible according to https://github.com/aquasecurity/trivy-action/issues/309.

mikehalmamoj (approved):

PR #326, opened 2024-04-29
SDIT-1681: 🐛 Fix slack alerts

petergphillips:


.github/workflows/security_trivy.yml
```diff
@@ -31,13 +31,13 @@ jobs:
         format: 'sarif'
         output: 'trivy-results.sarif'
         exit-code: 1
-        limit-severities-for-sarif: true
```
petergphillips, line 34: note that this will cause the trivy tests to fail and thus I can test the slack notification 😁

mikehalmamoj (approved):

PR #337, opened 2024-05-02
SDIT-1617 Switch to the new prison-api endpoint

mikehalmamoj:
Happy with last index refresh in preprod, so we're going for prod.


petergphillips (approved):
LGTM

PR #339, opened 2024-05-02
SDIT-1617 Add all offences to Prisoner record

mikehalmamoj:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Offence.kt
```diff
@@ -0,0 +1,28 @@
...
+  val offenceDate: LocalDate? = null,
+
+  @Schema(description = "Indicates offence is from the latest booking", example = "true")
+  val latestBooking: Boolean,
+
+  @Schema(description = "Indicates the offence is active", example = "true")
+  val active: Boolean = true,
+
+  @Schema(description = "Indicates the offence is convicted", example = "true")
+  val convicted: Boolean = true,
```
mikehalmamoj, line 27: Added these to indicate that we don't include all offences, only those which are active and convicted (as returned from prison-api). This was a suggestion from @andymarke to "future proof" us in case we ever need to add inactive or unconvicted offences.

petergphillips (approved):
LGTM

PR #340, opened 2024-05-03
SDIT-1617 Tidy up some remnants from the old prison-api endpoint

mikehalmamoj:


hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/HmppsDomainEventsEmitterIntTest.kt
```diff
@@ -209,7 +209,7 @@ class HmppsDomainEventsEmitterIntTest : IntegrationTestBase() {
                 prisonerNumber = "A1239DD",
                 bookingId = null,
                 firstName = "NEW_NAME",
-              ).toOffenderBookingOld(),
+              ).toOffenderBookingNewEndpoint(),
```
petergphillips, line 212: is it worth renaming `toOffenderBookingNewEndpoint` as well?

PR #341, opened 2024-05-07
Revert "SDIT-1617 Add all offences to Prisoner record (#339)"

mikehalmamoj:
This reverts commit 7ce52ab1d41dafb8c893bbd73e882d07f6c02277. Because we've realised that charge / court case status aren't accurate for historical offences, so we'll replace this with a new attribute on the Prisoner record called `convictedOffences`.


petergphillips (approved):
LGTM

PR #346, opened 2024-05-08
SDIT-1618 Refactor duplicated DTOs / translator

mikehalmamoj:
There was a lot of duplication for external DTOs in the indexer production code and search test code. And also with the Prisoner/OffenderBooking translator being used by both. So I kept forgetting to update the search test code and write tests in there 🤦! I've moved all the DTOs and the translator into the common project so they don't need to be kept in sync (which might remind me to write tests for search when I add new elements to the model).


andymarke (approved):
lgtm

PR #352, opened 2024-05-13
SDIT-1653 Add email addresses to the Prisoner record

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -60,7 +60,8 @@ private fun findAttributes(
         listOf(propertyName to Attribute(propClass, openSearchName, nested, fuzzy))
       }
     }
-    PropertyType.LIST -> {
+    PropertyType.LIST_SIMPLE -> throw InvalidAttributeTypeException("Attribute '${prefix}${prop.name}' is invalid. Lists of simple types are not supported because they cannot be extended (and would require special support from OpenSearch adding). Please change to a list of objects. See [Prisoner.EmailAddresses] for an example.")
```
mikehalmamoj, line 63: Considered making email addresses a `List<String>` but realised that this prevents us from adding further details to emails (e.g. primary email flag) if we get further requirements. Decided to prevent application startup if somebody tries to create a Prisoner attribute with a list of simple types.

petergphillips (approved):
LGTM

PR #360, opened 2024-05-16
SDIT-1698 Add DL, NINO and alias identifiers to the Prisoner record

mikehalmamoj:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/translator.kt
```diff
@@ -216,3 +219,14 @@ private fun String.extractNumbers() =
...
+private fun List<OffenderIdentifier>?.toIdentifiers(): List<Identifier>? =
+  this?.mapNotNull {
+    when (it.type) {
+      "PNC" -> Identifier("PNC", it.value.toPncNumber(), it.issuedDate, it.issuedAuthorityText, it.whenCreated.withNano(0))
+      "CRO", "DL", "NINO" -> Identifier(it.type, it.value, it.issuedDate, it.issuedAuthorityText, it.whenCreated.withNano(0))
+      else -> null
+    }
+  }?.sortedWith(compareBy<Identifier> { it.createdDateTime }.thenBy { it.type })
+
+private fun String.toPncNumber(): String = if (this.isPNCNumber()) this.canonicalPNCNumberShort()!! else this
```
mikehalmamoj, line 232: I decided to only store the short form PNC. This is because it's the most commonly used so should be recognisable by users (if they ever see it).

When we add search in for identifiers we'll convert any search terms to the same format.

andymarke (approved):
lgtm

PR #361, opened 2024-05-17
SDIT-1698 Add alias identifiers to Prisoner record

mikehalmamoj:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/nomis/OffenderBooking.kt
```diff
@@ -25,7 +26,7 @@ data class OffenderBooking(
   val physicalMarks: List<PhysicalMark>? = null,
   val csra: String? = null,
   val categoryCode: String? = null,
-  val identifiers: List<OffenderIdentifier>? = null,
+  val allIdentifiers: List<OffenderIdentifier>? = null,
```
mikehalmamoj, line 29: Replacing identifiers which only come from the active offender with allIdentifiers which includes all the alias identifiers.

petergphillips (approved):
LGTM

PR #369, opened 2024-05-30
SDIT-1754: ✨ Implement search of location event changes

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/SearchClient.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.opensearch.client.ResponseException
+import org.opensearch.client.RestHighLevelClient
+import org.opensearch.client.core.CountRequest
+import org.opensearch.client.core.CountResponse
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+@Service
+class SearchClient(
```
mikehalmamoj, line 16: Maybe consider this for moving to `common` once this solution is working OK?

mikehalmamoj (approved):

PR #384, opened 2024-07-05
SDIT-1860 Remove COURT_SENTENCE-CHANGED aka C_NOTIFICATION event

andymarke:
Other events already cover that NOMIS tables we are interested in, in particular OFFENDER_CHARGES-UPDATED


petergphillips (approved):
LGTM

PR #385, opened 2024-07-08
SDIT-1885: ✨ Experiment with locking down fields to allowed client fields only

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonerSearchService.kt
```diff
@@ -132,6 +133,11 @@ class PrisonerSearchService(
         size(RESULT_HITS_MAX)
         query(it)
       }
+      // TODO: remove the -[0-9]* suffix from the client
+      fields.clients[authenticationHolder.clientId]?.apply {
+        searchSourceBuilder.fetchSource(this.toTypedArray(), null)
```
petergphillips, line 138: this line ensures that only the specified fields are returned by OpenSearch

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonerSearchService.kt
```diff
@@ -132,6 +133,11 @@ class PrisonerSearchService(
         size(RESULT_HITS_MAX)
         query(it)
       }
+      // TODO: remove the -[0-9]* suffix from the client
+      fields.clients[authenticationHolder.clientId]?.apply {
```
petergphillips, line 137: eventually would expect that an exception is thrown if client fields are not defined for the given client id

hmpps-prisoner-search/src/main/resources/application-client-fields.yml
```diff
@@ -0,0 +1,11 @@
+search.clients:
+  restricted-patients-client:
```
petergphillips, line 2: list of fields obtained from restricted patients dto - https://github.com/ministryofjustice/hmpps-restricted-patients-api/blob/54cc87507b653d6ff6b81c0e778bf90f20f75712/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsrestrictedpatientsapi/model/response/PrisonerResult.kt#L6-L16

PR #386, opened 2024-07-12
CPR-346-add-role

JunaedAhmedMOJ:
created new role following hmpps role standards for core person records

### DIFF
--- a/hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonerSearchResourceTest.kt
+++ b/hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonerSearchResourceTest.kt
@@ -88,9 +88,9 @@ class PrisonerSearchResourceTest : AbstractSearchDataIntegrationTest() {
   }
 
   @Test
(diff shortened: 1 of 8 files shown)


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonerSearchResourceTest.kt
```diff
@@ -79,7 +79,16 @@ class PrisonerSearchResourceTest : AbstractSearchDataIntegrationTest() {
...
   @Test
-  fun `search by prisonId success for ROLE_GLOBAL_SEARCH and ROLE_PRISONER_SEARCH role`() {
+  fun `search by prisonId success for PRISONER_SEARCH__CORE_PERSON__RO role`() {
+    webTestClient.get().uri("/prisoner-search/prison/MDI")
+      .headers(setAuthorisation(roles = listOf("PRISONER_SEARCH__CORE_PERSON__RO")))
+      .header("Content-Type", "application/json")
+      .exchange()
+      .expectStatus().isOk
+  }
+
```
markreesmoj, line 89: This test isn't correct, we don't need to call this endpoint

andymarke (approved):
lgtm

PR #388, opened 2024-07-18
SDIT-1896 Fix bug with traces for sending messages direct to queues

mikehalmamoj:


hmpps-prisoner-search-indexer/src/test/resources/application-test.yml
```diff
@@ -47,27 +47,22 @@ hmpps.sqs:
       queueName: ${random.uuid}
       dlqName: ${random.uuid}
       dlqMaxReceiveCount: 1
-      propagateTracing: false
```
mikehalmamoj, line 50: This was a previous workaround to fix tests - not needed now because the bug is fixed

karenmillermoj (approved):
lgtm

PR #391, opened 2024-08-05
SDIT-1928 Investigate strategies for prisoner search selective updates

steverendell:
Part 1 - Multiple indices, i.e. a separate one for Incentives

### DIFF
--- a/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
+++ b/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
@@ -10,6 +10,7 @@ import org.springframework.data.elasticsearch.annotations.FieldType
 import java.time.LocalDateTime
 
 const val INDEX_STATUS_ID = "STATUS"
+const val INCENTIVES_INDEX_STATUS_ID = "INCENTIVES_STATUS"
 
 enum class IndexState(val active: Boolean) {
   ABSENT(false),
@@ -24,7 +25,7 @@ data class IndexStatus(
   @Id
   @Field(type = FieldType.Keyword)
   @JsonIgnore
-  val id: String = INDEX_STATUS_ID,
+  val id: String,
 
   @Field(type = FieldType.Keyword)
   @Schema(description = "The index currently active for searches", example = "GREEN")
@@ -58,7 +59,7 @@ data class IndexStatus(
 
   val otherIndex
     @Schema(description = "The index currently available for rebuilding", example = "BLUE")
-    get() = currentIndex.otherIndex()
+    get() = currentIndex.otherIndex(id)
 
   fun inProgress(): Boolean = this.otherIndexState == IndexState.BUILDING
 
@@ -112,6 +113,6 @@ data class IndexStatus(
   fun isNotBuilding() = isBuilding().not()
 
   companion object {
-    fun newIndex() = IndexStatus(currentIndex = SyncIndex.NONE)
+    fun newIndex(indexStatusId: String, currentIndex: SyncIndex) = IndexStatus(id = indexStatusId, currentIndex = currentIndex)
   }
 }
--- a/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerDifference.kt
+++ b/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerDifference.kt
@@ -28,3 +28,10 @@ internal fun getDiffResult(prisoner: Prisoner, other: Prisoner): DiffResult<Pris
       .filter { it.findAnnotations<DiffableProperty>().isNotEmpty() }
       .forEach { property -> append(property.name, property.call(prisoner), property.call(other)) }
   }.build()
+
+internal fun getDiffResult(prisoner: Incentive, other: Incentive): DiffResult<Incentive> =
+  DiffBuilder(prisoner, other, ToStringStyle.JSON_STYLE).apply {
+    Incentive::class.members
+      .filter { it.findAnnotations<DiffableProperty>().isNotEmpty() }
+      .forEach { property -> append(property.name, property.call(prisoner), property.call(other)) }
+  }.build()
--- a/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/SyncIndex.kt
+++ b/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/SyncIndex.kt
@@ -5,11 +5,28 @@ enum class SyncIndex(val indexName: String) {
   GREEN("prisoner-search-green"),
   BLUE("prisoner-search-blue"),
   NONE("new-indexes"),
+
+  GREEN_I("incentive-search-green"),
+  BLUE_I("incentive-search-blue"),
+  NONE_I("new-incentive-indexes"),
   ;
 
-  fun otherIndex(): SyncIndex = when (this) {
+  fun otherIndex(indexName: String): SyncIndex = when (this) {
     GREEN -> BLUE
     BLUE -> GREEN
     NONE -> GREEN
+
+    GREEN_I -> BLUE_I
+    BLUE_I -> GREEN_I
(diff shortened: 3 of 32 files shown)


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
```diff
@@ -10,6 +10,7 @@ import org.springframework.data.elasticsearch.annotations.FieldType
 import java.time.LocalDateTime
 
 const val INDEX_STATUS_ID = "STATUS"
+const val INCENTIVES_INDEX_STATUS_ID = "INCENTIVES_STATUS"
```
steverendell, line 13: Store a second document in the status index to manage the new incentives index

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/SyncIndex.kt
```diff
@@ -5,11 +5,28 @@ enum class SyncIndex(val indexName: String) {
   GREEN("prisoner-search-green"),
   BLUE("prisoner-search-blue"),
   NONE("new-indexes"),
+
+  GREEN_I("incentive-search-green"),
+  BLUE_I("incentive-search-blue"),
+  NONE_I("new-incentive-indexes"),
```
steverendell, line 11: 3 new values to reflect incentives state. Could improve names later.
I tried a second SyncIndex class with inheritance but hit a problem deserialising the status index's currentIndex value from a string to the appropriate class.

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
```diff
@@ -29,7 +30,7 @@ class PrisonerRepository(
   }
 
   fun count(index: SyncIndex) = try {
-    client.count(CountRequest(index.indexName), RequestOptions.DEFAULT).count
+    client.count(CountRequest(index.indexName()), RequestOptions.DEFAULT).count
```
steverendell, line 33: This indexName() fun is probably redundant but left it in for now.

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/CompareIndexService.kt
```diff
@@ -37,7 +38,7 @@ class CompareIndexService(
 
   fun doIndexSizeCheck(): SizeCheck {
     val start = System.currentTimeMillis()
-    val totalIndexNumber = prisonerRepository.count(indexStatusService.getIndexStatus().currentIndex)
+    val totalIndexNumber = prisonerRepository.count(indexStatusService.getIndexStatus(INDEX_STATUS_ID).currentIndex)
```
steverendell, line 41: In order to generalise various index service classes the index id now needs to be specified

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/IncentiveRepository.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.data.elasticsearch.core.ElasticsearchOperations
+import org.springframework.data.elasticsearch.core.mapping.IndexCoordinates
+import org.springframework.data.elasticsearch.core.query.IndexQueryBuilder
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.prisonersearch.common.config.OpenSearchIndexConfiguration
+import uk.gov.justice.digital.hmpps.prisonersearch.common.model.INCENTIVES_INDEX_STATUS_ID
+import uk.gov.justice.digital.hmpps.prisonersearch.common.model.Incentive
+import uk.gov.justice.digital.hmpps.prisonersearch.common.model.SyncIndex
+
```
steverendell, line 22: Very similar to PrisonerRepository but handles the Incentive document class, index and GREEN/BLUE states

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
```diff
@@ -112,6 +113,6 @@ data class IndexStatus(
   fun isNotBuilding() = isBuilding().not()
 
   companion object {
-    fun newIndex() = IndexStatus(currentIndex = SyncIndex.NONE)
+    fun newIndex(indexStatusId: String, currentIndex: SyncIndex) = IndexStatus(id = indexStatusId, currentIndex = currentIndex)
```
steverendell, line 116: newIndex now needs to be passed both the index id and initial state

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
mikehalmamoj: Why would the prisoner / incentive indexes need managing separately? Wouldn't they come as a "set" and all be doing the same thing at the same time?

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerDifference.kt
```diff
@@ -28,3 +28,10 @@ internal fun getDiffResult(prisoner: Prisoner, other: Prisoner): DiffResult<Pris
       .filter { it.findAnnotations<DiffableProperty>().isNotEmpty() }
       .forEach { property -> append(property.name, property.call(prisoner), property.call(other)) }
   }.build()
+
+internal fun getDiffResult(prisoner: Incentive, other: Incentive): DiffResult<Incentive> =
+  DiffBuilder(prisoner, other, ToStringStyle.JSON_STYLE).apply {
+    Incentive::class.members
+      .filter { it.findAnnotations<DiffableProperty>().isNotEmpty() }
+      .forEach { property -> append(property.name, property.call(prisoner), property.call(other)) }
+  }.build()
```
mikehalmamoj, line 37: If we run with this solution this function could be genericised on type T.

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/SyncIndex.kt
mikehalmamoj: If all indexes were tied together instead of managed independently then would `val indexName: String` become `val indexNames: List<String>`?

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
steverendell: I wasnt sure how flexible we need to be on this, e.g. is the ability to swap from the A to the B index useful for one of the indices? 🤷

mikehalmamoj: Yeah I'm just thinking it through some more... it might actually be quite useful to manage them independently. For example if one has its model changed you only need to rebuild that one. 

Or maybe it would just get too complicated - e.g. NOMIS data is GREEN, incentives is BLUE and GREEN is currently rebuilding, restricted patients is GREEN but BLUE is rebuilding. I'm lost now.

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/SyncIndex.kt
steverendell: Yep, or I could add a 'incentiveIndexName' etc so the same 3 enums could be used for all indices and they could remain independent

PR #393, opened 2024-08-08
SDIT-1930: ✨ Latest sqs and add in eventType for sent messages

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/IndexQueueService.kt
```diff
@@ -52,7 +53,11 @@ class IndexQueueService(
   private fun sendMessage(request: IndexMessageRequest): SendMessageResponse = indexSqsClient.sendMessage(
     SendMessageRequest.builder().queueUrl(indexQueueUrl)
       .messageBody(objectMapper.writeValueAsString(request))
-      .build(),
+      .messageAttributes(
+        mapOf(
+          "eventType" to MessageAttributeValue.builder().dataType("String").stringValue("hmpps-prisoner-search-indexer-${request.type?.name?.lowercase()}").build(),
```
petergphillips, line 58: these aren't really necessary, but aid logging in that instead of the app insights event (dependency or request) being just a `PUBLISH` or `RECEIVE` they contain the event type as well

andymarke (approved):
lgtm

PR #395, opened 2024-08-12
SDIT-1933 Raise "receive" event for move booking

andymarke:
Fix issue where integration tests fire receive event when recreating prisoners


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
```diff
@@ -158,6 +158,7 @@ private fun Prisoner.isNewAdmission(previousPrisonerSnapshot: Prisoner?) =
 
 private fun Prisoner.isNewAdmissionDueToMoveBooking(previousPrisonerSnapshot: Prisoner?) =
   previousPrisonerSnapshot?.bookingId == null &&
+    this.bookingId != null &&
     this.status == "ACTIVE IN"
 
 private fun Prisoner.isReadmission(previousPrisonerSnapshot: Prisoner?) =
```
andymarke, line 164: A hack to get integration tests to work.

Obviously in real world if you have a active booking there must be bookingId - so the extra statement is redundant but does no harm but this will prevent extra domain events being raised during the creation of tests data that is breaking a lot of tests.

I will look at a better stable solution to integration tests since they are getting very messy. But for now now this will fix `main`

petergphillips (approved):
LGTM

PR #398, opened 2024-08-13
SDIT-1934 Fixed integration test and removes hack for move booking

andymarke:
Most integration tests were accidentally setting the bookingId to null even though the data had a booking, this caused additionally domain events throwing off the tests.

So removed hack in code to ignore scenario and corrected the seed data

### DIFF
--- a/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
+++ b/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
@@ -158,7 +158,6 @@ private fun Prisoner.isNewAdmission(previousPrisonerSnapshot: Prisoner?) =
 
 private fun Prisoner.isNewAdmissionDueToMoveBooking(previousPrisonerSnapshot: Prisoner?) =
   previousPrisonerSnapshot?.bookingId == null &&
-    this.bookingId != null &&
     this.status == "ACTIVE IN"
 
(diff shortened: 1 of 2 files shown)


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
```diff
@@ -158,7 +158,6 @@ private fun Prisoner.isNewAdmission(previousPrisonerSnapshot: Prisoner?) =
 
 private fun Prisoner.isNewAdmissionDueToMoveBooking(previousPrisonerSnapshot: Prisoner?) =
   previousPrisonerSnapshot?.bookingId == null &&
-    this.bookingId != null &&
     this.status == "ACTIVE IN"
```
andymarke, line 162: bookingId can not be null given the status is ACTIVE IN - this was just hack to get the tests to pass

hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/HmppsDomainEventsEmitterIntTest.kt
```diff
@@ -196,7 +196,7 @@ class HmppsDomainEventsEmitterIntTest : IntegrationTestBase() {
 
   @Test
   fun `e2e - will send prisoner updated event to the domain topic`() {
```
andymarke, line 198: all these tests were seeding a booking with a null bookingId that made no sense and caused the receive domain event to fire accidentally

mikehalmamoj (approved):

PR #415, opened 2024-09-11
SDIT-2002 Use  hmppsDomainTopic.publish

andymarke:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventPublisherListener.kt
```diff
@@ -19,24 +19,17 @@ class DomainEventPublisherListener(
...
-          "eventType" to MessageAttributeValue.builder().dataType("String").stringValue(event.eventType).build(),
-        ),
-      ).build()
-
-    topicSnsClient.publish(request).join()
+    hmppsDomainTopic.publish(
+      eventType = event.eventType,
+      event = event.body,
+      retryPolicy = NeverRetryPolicy(),
+    )
```
mikehalmamoj, line 32: Are you sure that this shouldn't retry? If the openSearch update succeeds then when the prisoner event message is retried it will find that nothing has changed - so another domain event won't be published.

mikehalmamoj (approved):

PR #417, opened 2024-09-13
SDIT-2041 Retry get incentive level a few times if it fails

andymarke:
When a failure happens, particularly after a new booking, multiple receive domain events can be raised since we fall back to using previous incentive level - but that value will typically be incorrect for a new booking so "differences" are incorrectly found.

This mitigates by trying a few times before failing.


petergphillips (approved):
LGTM

PR #440, opened 2024-10-23
SDIT-2123: 🔊 Reduce dependency sampling to 5% and change logging to info level

petergphillips:


hmpps-prisoner-search-indexer/applicationinsights.dev.json
```diff
@@ -16,16 +16,38 @@
   "sampling": {
     "percentage": 100,
     "overrides": [
+      {
+        "telemetryType": "dependency",
+        "attributes": [
+          {
+            "key": "server.address",
+            "value": ".+",
```
petergphillips, line 24: we want to keep `InProc` `PUBLISH xxx` events at 100%, but want to reduce everything else.  Can't seem to find a good way to do it, so trying this way instead

hmpps-prisoner-search/applicationinsights.dev.json
```diff
@@ -16,16 +16,31 @@
   "sampling": {
     "percentage": 100,
     "overrides": [
+      {
+        "telemetryType": "dependency",
```
petergphillips, line 20: this reduces all dependencies to 5%

hmpps-prisoner-search-indexer/applicationinsights.dev.json
mikehalmamoj: Hopefully you will eventually find a better way.

mikehalmamoj (approved):

PR #443, opened 2024-10-25
SDIT-2077 Unnecessary exception logging noise needs removing

steverendell:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/OffenderEventListener.kt
```diff
@@ -85,6 +86,20 @@ class OffenderEventListener(
 
         else -> log.warn("We received a message of event type {} which I really wasn't expecting", eventType)
       }
+    } catch (olfe: OptimisticLockingFailureException) {
+      if (true == olfe.message?.contains("Cannot index a document due to seq_no+primary_term conflict")) {
```
petergphillips, line 90: I'm not a great fan of writing `if` statements in reverse

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/OffenderEventListener.kt
```diff
@@ -85,6 +86,20 @@ class OffenderEventListener(
...
+        // This is not an error and so we want to avoid exceptions being logged
+        val (message, _, messageAttributes) = fromJson<Message>(requestJson)
+        log.info("Detected a seq_no+primary_term conflict and trying again for message:\n$message")
```
petergphillips, line 93: better to put the `message` as a parameter and not use string interpolation

petergphillips (approved):
LGTM

PR #448, opened 2024-11-04
SDIT-2102 Add RED index to the refresh from prod - fix

steverendell:
Fix possible timeout due to longer restore time


petergphillips (approved):
LGTM

PR #456, opened 2024-11-20
SDIT-: ♻️ Move prisoner removed event to listener service

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -215,7 +213,6 @@ class PrisonerSynchroniserService(
   fun delete(prisonerNumber: String) =
     prisonerRepository.delete(prisonerNumber).also {
       prisonerRepository.delete(prisonerNumber, SyncIndex.RED)
-      hmppsDomainEventEmitter.emitPrisonerRemovedEvent(prisonerNumber)
```
petergphillips, line 218: turns out that this method is also called for `BOOKING_NUMBER-CHANGED` events, which we don't want to then generate a prisoner removed event for here.

andymarke (approved):
lgtm

PR #461, opened 2024-12-03
HMAI-66 Add Date of birth to prisoner details

ben-madley-mt:
Add date of birth as a filter field into the `prisoner-details` endpoint. This will allow HMPPS Integration API to use a search that filters by Prison ID. As a bonus, this increases the parity with the `global-search` API.


petergphillips (approved):
LGTM

PR #467, opened 2025-01-13
SDIT-2182 Create duplicate logging from the red index updates - handle create

steverendell:
prisoner creation in opensearch needs to check it does not already exist

### DIFF
--- a/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
+++ b/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
@@ -19,6 +19,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.data.elasticsearch.core.ElasticsearchOperations
 import org.springframework.data.elasticsearch.core.document.Document
 import org.springframework.data.elasticsearch.core.mapping.IndexCoordinates
+import org.springframework.data.elasticsearch.core.query.IndexQuery
 import org.springframework.data.elasticsearch.core.query.IndexQueryBuilder
 import org.springframework.data.elasticsearch.core.query.ScriptType
 import org.springframework.data.elasticsearch.core.query.UpdateQuery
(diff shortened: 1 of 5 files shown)


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
```diff
@@ -59,6 +60,19 @@ class PrisonerRepository(
...
       .toPrisonerDocumentSummary(prisonerNumber)
 
+  fun createPrisoner(prisoner: Prisoner, index: SyncIndex) {
+    val response = openSearchRestTemplate.index(
+      IndexQueryBuilder()
+        .withObject(prisoner)
+        .withOpType(IndexQuery.OpType.CREATE)
+        .build(),
+      index.toIndexCoordinates(),
+    )
```
steverendell, line 70: This throws an exception if the prisoner already exists in the index

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -116,19 +115,22 @@ class PrisonerSynchroniserService(
...
 
       val prisoner = Prisoner().translate(
         existingPrisoner = null,
         ob = ob,
         incentiveLevel = runCatching { getIncentive(ob) },
         restrictedPatientData = runCatching { getRestrictedPatient(ob) },
       )
-      prisonerRepository.save(prisoner, SyncIndex.RED)
+      prisonerRepository.createPrisoner(prisoner, SyncIndex.RED)
+      // If prisoner already exists in opensearch, an exception is thrown (same as for version conflict with update)
```
steverendell, line 129: So any multiple attempts to create, due to multiple sqs events will fail with an exception casing a retry. Which will then do an update instead and _probably_ result in a no-op.

mikehalmamoj (approved):

PR #474, opened 2025-01-24
SDIT-2489-Manual-reindex-of-prisoner-does-not-update-red-index

steverendell:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/MaintainIndexService.kt
```diff
@@ -186,21 +186,27 @@ class MaintainIndexService(
...
-    nomisService.getOffender(prisonerNumber)?.let { ob ->
+  private fun sync(prisonerNumber: String, activeIndices: List<SyncIndex>): Prisoner {
+    val offenderBooking = nomisService.getOffender(prisonerNumber)
+    return offenderBooking?.let { ob ->
       prisonerSynchroniserService.reindex(ob, activeIndices, "MAINTAIN")
     }
+      ?.also {
+        prisonerSynchroniserService.reindexUpdate(offenderBooking, "MAINTAIN")
+        prisonerSynchroniserService.reindexIncentive(prisonerNumber, SyncIndex.RED, "MAINTAIN")
+        prisonerSynchroniserService.reindexRestrictedPatient(prisonerNumber, offenderBooking, SyncIndex.RED, "MAINTAIN")
```
steverendell, line 197: Cannot just do a save() as we need the telemetry.
This is a bit wasteful as it gets the document and incentive data more than once but is temporary. Once we have switched indexes it can be simplified.

PR #475, opened 2025-01-24
SDIT-2182 Create duplicate logging from the red index updates - handle delete

steverendell:
Creation of an event was not simulated for prisoner / alias deletion.
Also make the incentive code field non-null.


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/IndexListenerService.kt
```diff
@@ -73,6 +73,7 @@ class IndexListenerService(
         log.debug("Delete check: offender ID {} no longer exists, deleting", this)
         prisonerSynchroniserService.delete(prisonerNumber = this)
         hmppsDomainEventEmitter.emitPrisonerRemovedEvent(offenderNo = this, red = false)
+        hmppsDomainEventEmitter.emitPrisonerRemovedEvent(offenderNo = this, red = true)
```
steverendell, line 76: This looks a little odd but is because the delete() call does all the indices. One of these event calls emits an event , the other simulates for the red index

andymarke (approved):
lgtm

PR #485, opened 2025-02-20
SDIT-2604 Updated event generated on red index when nothing has changed

steverendell:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -88,10 +88,7 @@ class PrisonerSynchroniserService(
           log.warn("Prisoner not found in RED index for {}", ob.offenderNo)
           throw PrisonerNotFoundException(ob.offenderNo)
         }
-        prisonerDifferenceService.getDifferencesByCategory(summary.prisoner, prisoner)
-          .also {
-            domainEventEmitter.emitPrisonerDifferenceEvent(ob.offenderNo, it, red = true)
-          }
+        prisonerDifferenceService.generateDiffEvent(summary.prisoner, ob.offenderNo, prisoner, true)
```
steverendell, line 91: When the db hash has changed, a difference event is generated but only if there is at least one ‘diffable property’ change.
The red index code here was missing this check, generating an event whenever the prisoner record had changed even when there are no diffable property changes. generateDiffEvent() includes the check

andymarke (approved):
lgtm

PR #493, opened 2025-03-05
SDIT-2619-Switch-domain-event-generation-to-use-the-new-RED-index

steverendell:


hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/MatchDomainEvents.kt
```diff
@@ -0,0 +1,190 @@
+package uk.gov.justice.digital.hmpps.prisonersearch.indexer
+
```
steverendell, line 2: Included this temporary event checking program.

andymarke (approved):
LGTM

PR #497, opened 2025-03-11
SDIT-2647-Trap-failure-to-send-domain-event

steverendell:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/HmppsDomainEventEmitter.kt
```diff
@@ -91,36 +80,23 @@ class HmppsDomainEventEmitter(
     val prisonerUpdatedEvent = PrisonerUpdatedEvent(offenderNo, differences.keys.toList().sorted())
     val event = PrisonerUpdatedDomainEvent(prisonerUpdatedEvent, Instant.now(clock), diffProperties.host)
     if (red) {
-      event.publish {
-        log.error("Failed to send event $UPDATED_EVENT_TYPE for offenderNo=$offenderNo, differences=$differences. Event will be retried")
-        throw it
-      }
+      event.publish { logError(event, it, differences) }
```
steverendell, line 83: The main difference is that now there is no point in re-throwing the exception as the feature of it causing a db rollback no longer applies. So the red index update is not rolled back and a retry would just result in a no-op and therefore no event generation. 
So the strategy now is to alert on the EVENTS_SEND_FAILURE and intervene manually. This is ok as the failure is extremely rare (no occurrences in 2025 anyway)

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/HmppsDomainEventEmitter.kt
```diff
@@ -247,15 +235,6 @@ open class PrisonerDomainEvent<T : PrisonerAdditionalInformation>(
...
 
-data class PersonReference(val identifiers: List<Identifier>) {
-  companion object {
-    private const val NOMS_NUMBER_TYPE = "NOMS"
-    fun withNomsNumber(prisonNumber: String) = PersonReference(listOf(Identifier(NOMS_NUMBER_TYPE, prisonNumber)))
-  }
-
-  data class Identifier(val type: String, val value: String)
-}
-
```
steverendell, line 258: Not used

hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/OffenderEventListenerIntTest.kt
```diff
@@ -62,6 +62,7 @@ class OffenderEventListenerIntTest : IntegrationTestBase() {
     await untilAsserted {
       val prisoner = prisonerRepository.get(prisonerNumber, listOf(SyncIndex.RED))
       assertThat(prisoner?.prisonerNumber).isEqualTo(prisonerNumber)
+      assertThat(getNumberOfMessagesCurrentlyOnDomainQueue()).isEqualTo(2)
```
steverendell, line 65: Added some extra checks to tighten up some of the tests

andymarke (approved):
LGTM

PR #503, opened 2025-03-14
SDIT-2620 Remove old index code - part 2

steverendell:
Remove all mention of specific indexes. Only 1 default index now used.
The build functionality still works and refreshes the index without triggering events
The refresh functionality does the same but *does* trigger events


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/IndexStatus.kt
```diff
@@ -12,7 +12,6 @@ import java.time.LocalDateTime
 const val INDEX_STATUS_ID = "STATUS"
 
 enum class IndexState(val active: Boolean) {
-  ABSENT(false),
```
steverendell, line 15: I kept this class as it still keeps track of whether a build is in progress, even though it is only refreshing

hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/repository/PrisonerRepository.kt
```diff
@@ -26,56 +25,37 @@ class PrisonerRepository(
...
+  fun get(prisonerNumber: String): Prisoner? = openSearchRestTemplate.get(prisonerNumber, Prisoner::class.java, of(PRISONER_INDEX))
 
-  fun createIndex(index: SyncIndex) {
-    log.info("creating index {}", index.indexName)
-    client.indices().create(CreateIndexRequest(index.indexName), RequestOptions.DEFAULT)
-    addMapping(index)
+  fun createIndex() {
+    log.info("creating index")
+    client.indices().create(CreateIndexRequest(PRISONER_INDEX), RequestOptions.DEFAULT)
+    addMapping()
```
steverendell, line 44: create and delete still needed for testing

andymarke (approved):
lgtm

PR #506, opened 2025-03-24
SDIT-2681-Add-emergency-fix-endpoints-for-domain-events

steverendell:
This covers some of the additional events which are no longer retried automatically when a publish call fails.


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/resource/DomainEventsResource.kt
```diff
@@ -51,6 +51,73 @@ class DomainEventsResource(private val domainEventEmitter: HmppsDomainEventEmitt
       occurredAt = occurredAtTime,
     )
   }
+
+  @PutMapping("/events/prisoner/released/{prisonerNumber}")
+  @ResponseStatus(HttpStatus.ACCEPTED)
```
petergphillips, line 56: should this be `ACCEPTED`?  Would `NO_CONTENT` be more appropriate since the event is sent immediately?

petergphillips (approved):
LGTM

PR #511, opened 2025-03-28
SDIT-2177-Move-alerts-to-domain-partial-update-2

steverendell:
- Switch to person.alerts.changed event
- Impose explicit order on alert list

### DIFF
--- a/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventListener.kt
+++ b/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventListener.kt
@@ -33,13 +33,6 @@ class DomainEventListener(
       "restricted-patients.patient.removed",
       "restricted-patients.patient.supporting-prison-changed",
     )
-    private val alertEvent = setOf(
-      "person.alert.created",
-      "person.alert.updated",
-      "person.alert.deleted",
-      // "person.alert.inactive", this event is not handled as it only occurs alongside a "person.alert.updated" event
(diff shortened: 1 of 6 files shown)


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventListener.kt
```diff
@@ -52,7 +45,7 @@ class DomainEventListener(
       when (eventType) {
         in incentiveEvent -> indexListenerService.incentiveChange(fromJson(message), eventType)
         in restrictedPatientEvent -> indexListenerService.restrictedPatientChange(fromJson(message), eventType)
-        in alertEvent -> indexListenerService.alertChange(fromJson(message), eventType)
+        "person.alerts.changed" -> indexListenerService.alertChange(fromJson(message), eventType)
```
steverendell, line 48: This is more effeceint, and also fixes the Nomis batch problem with an alert being made inactive

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/AlertsService.kt
```diff
@@ -27,6 +27,8 @@ class AlertsService(
         .queryParam("isActive", true)
         .queryParam("page", 0)
         .queryParam("size", 1000)
+        .queryParam("sort", "activeFrom,DESC")
+        .queryParam("sort", "createdAt,ASC")
         .build(offenderNo)
```
steverendell, line 32: This will need a build kicking off as soon as it goes live

petergphillips (approved):
LGTM

PR #514, opened 2025-04-04
SDIT-2691 Support requesting which Prisoner attributes are returned in the response

mikehalmamoj:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonersInPrisonResource.kt
```diff
@@ -127,20 +134,34 @@ class PrisonersInPrisonResource(private val searchService: PrisonersInPrisonServ
     @RequestParam(value = "incentiveLevelCode", required = false)
     @Parameter(description = "Filter for the prisoners on an incentive level.", example = "STD")
     incentiveLevelCode: String?,
+    @RequestParam(value = "requestedAttributes", required = false)
+    @Parameter(description = "A list of attributes to return on the Prisoner record. An empty list defaults to all attributes.", example = "prisonerNumber,firstName")
+    requestedAttributes: List<String>? = null,
```
mikehalmamoj, line 139: The main change is to accept a list of attributes to populate on the returned `Prisoner` record.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -17,12 +17,33 @@ typealias Attributes = Map<String, Attribute>
...
+  }
+  .toCollection(LinkedHashSet())
+  .toList()
+
 /**
  * Derive all attributes that can be searched for in queries
  */
-internal fun getAttributes(kClass: KClass<*>): Attributes = kClass.memberProperties.flatMap { prop -> findAttributes(prop) }.toMap()
+internal fun getAttributes(kClass: KClass<*>): Attributes =
+  kClass.memberProperties.flatMap { prop -> findAttributes(prop) }.toMap()
```
mikehalmamoj, line 46: Anything else in this class is just ktlint doing its thing

common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Address.kt
```diff
@@ -8,20 +8,20 @@ import java.time.LocalDate
 
 data class Address(
   @Schema(description = "The full address on a single line.  No fixed address records will have the fullAddress set to 'No fixed address'.", example = "1 Main Street, Crookes, Sheffield, South Yorkshire, S10 1BP, England")
-  val fullAddress: String,
+  val fullAddress: String? = null,
```
mikehalmamoj, line 11: We have to make sure every single attribute on the Prisoner record is nullable in order to return a subset of the attributes.

common/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+
+class PrisonerTest {
+  @Test
+  fun `all Prisoner fields are nullable`() {
+    assertDoesNotThrow {
+      jacksonObjectMapper().readValue("{}", Prisoner::class.java)
+    }
+  }
```
mikehalmamoj, line 13: These tests check that all Prisoner attributes are nullable, including all nested objects

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -15,10 +15,32 @@ typealias Attributes = Map<String, Attribute>
 
 @Configuration
 class AttributeResolver {
+  // attributes contains everything you can use in queries in the attributes search
   @Bean
   fun attributes(): Attributes = getAttributes(Prisoner::class)
+
+  // attributesAndObjects contains everything you can request to be populated on the Prisoner record
+  @Bean
+  fun attributeAndObjects(attributes: Attributes): List<String> = attributes.attributeAndObjects()
```
mikehalmamoj, line 24: We have an endpoint that exposes the contents of `attributes()` to tell people using attributes search which attributes they can query over.

We'll need a similar endpoint exposing `attributesAndObjects()` to tell people which attributes they can request are returned from the Prisoner record.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/AtttributeResolver.kt
```diff
@@ -42,12 +64,24 @@ private fun findAttributes(
...
   PropertyType.LIST_OBJECTS -> {
     prop.getGenericTypeClass().memberProperties
-      .flatMap { childProp -> findAttributes(childProp, "${prefix}${prop.name}.", nested || prop.hasFieldAnnotation("Nested")) }
+      .flatMap { childProp ->
+        findAttributes(
+          childProp,
+          "${prefix}${prop.name}.",
+          nested || prop.hasFieldAnnotation("Nested"),
+        )
+      }
```
mikehalmamoj, line 73: ktlint

common/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerTest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+
+class PrisonerTest {
+  @Test
+  fun `all Prisoner fields are nullable`() {
+    assertDoesNotThrow {
+      jacksonObjectMapper().readValue("{}", Prisoner::class.java)
```
petergphillips, line 11: bit naff to instantiate this each time and might be different from the one created by spring.  Would be better to use `@JsonTest` and then you'll get the correct one injected.  See https://docs.spring.io/spring-boot/reference/testing/spring-boot-applications.html#testing.spring-boot-applications.json-tests.  Even if you don't use the `JacksonTester` it will still give you an object mapper.

common/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/PrisonerTest.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  @Test
+  fun `all PrisonerAlias fields are nullable`() {
```
petergphillips, line 16: this would be better to be parameterised, although would be even better to use recursive reflection to find out the objects, otherwise if someone adds another object then they'd need to remember to add it to this test class.

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonersInPrisonResource.kt
```diff
@@ -127,20 +134,32 @@ class PrisonersInPrisonResource(private val searchService: PrisonersInPrisonServ
     @RequestParam(value = "incentiveLevelCode", required = false)
     @Parameter(description = "Filter for the prisoners on an incentive level.", example = "STD")
     incentiveLevelCode: String?,
+    @RequestParam(value = "requestedAttributes", required = false)
```
petergphillips, line 137: not a great fan of this name.  It is a parameter on the request but actually operates on the response, hence including `requested` could lead to confusion.  Maybe `responseAttributes` would be better?
It is interesting that you are calling them attributes too.  Since it is json that is being returned I would have referred to them as fields or properties in the json rather than attributes of the json.

mikehalmamoj: Or maybe `responseFields` is actually better because then it's distinct from attribute search and might avoid confusion. They are 2 different thing after all, just sort of related 🤔

hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonersInPrisonResource.kt
```diff
@@ -127,20 +134,35 @@ class PrisonersInPrisonResource(private val searchService: PrisonersInPrisonServ
     @RequestParam(value = "incentiveLevelCode", required = false)
     @Parameter(description = "Filter for the prisoners on an incentive level.", example = "STD")
     incentiveLevelCode: String?,
+    @RequestParam(value = "responseFields", required = false)
+    @Parameter(
+      description = "A list of fields to populate on the Prisoner record returned in the response. An empty list defaults to all fields.",
+      example = "prisonerNumber,firstName,aliases.firstName,currentIncentive.level.code",
```
petergphillips, line 140: is this the correct format for an example list of strings - doesn't need brackets or anything then?

petergphillips (approved):
LGTM

PR #517, opened 2025-04-08
SDIT-2646 No fixed address flag should default to null

mikehalmamoj:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/translator.kt
```diff
@@ -257,6 +257,7 @@ private fun NomisAddress.toAddress(): Address {
     startDate = startDate,
     primaryAddress = primary,
     phoneNumbers = phones?.toPhoneNumbers(),
+    noFixedAddress = false,
```
mikehalmamoj, line 260: see line 222 where it is already set to true

petergphillips (approved):
LGTM

PR #518, opened 2025-04-08
SDIT-2646 Add responseFields parameter to the find by prison endpoint

mikehalmamoj:


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/PrisonerResponseFieldsTest.kt
```diff
@@ -247,4 +247,200 @@ class PrisonerResponseFieldsTest : AbstractSearchIntegrationTest() {
...
+      )
+
+      with(response.content.first()!!) {
+        assertThat(prisonerNumber).isEqualTo("A1234AA")
+        assertThat(dateOfBirth).isEqualTo("1975-07-20")
+        assertThat(recall).isEqualTo(true)
+        assertThat(heightCentimetres).isEqualTo(180)
+        assertThat(currentIncentive?.level?.description).isEqualTo("Standard")
+      }
+    }
```
mikehalmamoj, line 286: I've just copied all the same tests from the other endpoint I applied this change to. I don't think every endpoint needs all of these tests because they're essentially testing the same thing - that we can request objects, nested fields etc.

However I'm implementing all tests here just for expediency, will probably remove duplicate when applying to further endpoints.

andymarke (approved):
lgtm

PR #520, opened 2025-04-09
SDIT-2646 active should not be available to attribute search or response fields

mikehalmamoj:


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/ResponseFieldsResourceTest.kt
```diff
@@ -84,6 +84,13 @@ class ResponseFieldsResourceTest : IntegrationTestBase() {
         )
       }
     }
+
+    @Test
+    fun `should not return the derived type active`() {
+      webTestClient.getResponseFields { result: List<String> ->
+        assertThat(result).doesNotContain("active")
+      }
+    }
```
mikehalmamoj, line 93: I noticed that `Prisoner.active` is available in attribute search (but you can't use it for search) and response fields (but it isn't returned on the Prisoner record).

So this PR makes active unavailable.

petergphillips (approved):
LGTM

PR #522, opened 2025-04-28
SDIT-2733: ⬆️  Upgrade hmpps spring boot plugin to 8.1.0

petergphillips:


.circleci/config.yml
```diff
@@ -12,6 +12,7 @@ workflows:
     jobs:
       - path-filtering/filter:
           name: check-for-changes
+          tag: '3.13'
```
petergphillips, line 15: odd change this - looks like circleci have broken something recently and can't get the project to build using python 3.8 any more.

steverendell (approved):

PR #528, opened 2025-05-12
SDIT-2724-Add-admission-date-and-complexity-of-needs-fields-to-prisoner-search-document---CNL-performance-fix

steverendell:
Use prison register and OUT status to limit complexity-of-need calls


andymarke (approved):
LGTM

PR #537, opened 2025-05-20
SDIT-2724-Add-admission-date-and-complexity-of-needs-fields-to-prisoner-search-document---Cache-fix

steverendell:
Cache should expire, not be indefinite


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
```diff
@@ -171,10 +171,10 @@ class PrisonerRepository(
       client.lowLevelClient.performRequest(request)
       return true
     } catch (e: ResponseException) {
-      log.error("Unexpected response ${e.response} from delete of $prisonerNumber", e)
       if (e.response.statusLine.statusCode == 404) {
         return false
       }
+      log.error("Unexpected response ${e.response} from delete of $prisonerNumber", e)
```
steverendell, line 177: small tweak avoiding an incorrect exception log when the prisoner is just not in the index

PR #570, opened 2025-06-23
SDIT-2737 Ensure release name specific to indexer or search

karenmillermoj:


.github/actions/slack_release_results/action.yml
```diff
@@ -0,0 +1,90 @@
...
+    with:
+      channel-id: ${{ inputs.channel_id }}
+      payload: |
+          {
+            "blocks": [
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "*${{ github.inputs.k8s_deployment_name }}* version `${{ inputs.app_version }}` deploy to *${{ inputs.environment }}*"
```
karenmillermoj, line 63: This is the change from github actions  (now uses k8s_deployment_name, rather than  github.event.repository.name )

PR #583, opened 2025-07-21
SDIT-2671: ♻️  Use restricted patients generated model

petergphillips:


hmpps-prisoner-search-indexer/build.gradle.kts
```diff
@@ -83,3 +94,120 @@ tasks {
     compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_21
   }
 }
+
+data class ModelConfiguration(val name: String, val packageName: String, val url: String, val models: String = "") {
```
petergphillips, line 98: copied from to-nomis

PR #588, opened 2025-07-23
SDIT-2671: ♻️  Start to use prison api model - offender identifiers

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/nomis/OffenderBooking.kt
```diff
@@ -52,7 +53,7 @@ data class OffenderBooking(
   fun latestIdentifier(type: String) = allIdentifiers
     ?.filter { it.type == type }
     ?.takeIf { it.isNotEmpty() }
-    ?.maxBy { it.whenCreated }
+    ?.maxBy { it.whenCreated!! }
```
petergphillips, line 56: nomis model has `whenCreated` as optional (as generated from java), but we know it is mandatory

andymarke (approved):
lgtm

PR #590, opened 2025-07-23
SDIT-2671: ♻️  More prison api model (2)

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/translator.kt
```diff
@@ -140,11 +140,11 @@ fun Prisoner.translate(
 
   this.languages = ob.languages?.map {
     Language(
-      it.type,
+      it.type.value,
```
petergphillips, line 143: now enums so need `value`

steverendell (approved):
lgtm

PR #596, opened 2025-07-28
SDIT-2976: 👷 Remove compile warnings by adding param property

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/PrisonerListCriteria.kt
```diff
@@ -23,8 +20,6 @@ sealed class PrisonerListCriteria<out T> {
     val prisonerNumbers: List<String>,
   ) : PrisonerListCriteria<String>() {
 
-    override fun isValid() = prisonerNumbers.isNotEmpty() && prisonerNumbers.size <= 1000
```
petergphillips, line 26: the data class annotations above just had `@NotEmpty` etc. which meant previously they were applied only to the parameter.  Now they are also applied to the field which means that validation is triggered.  Note the same effect previously could have been achieved by changing the annotation to `@field:NotEmpty`

PR #618, opened 2025-08-14
pgp SDIT 2993 atlassian validation

petergphillips:
- **SDIT-2993: ✅ Try out using atlassian validator**
- **SDIT-2993: ✅ Try out using atlassian validator (2)**

### DIFF
--- a/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/OpenApiConfiguration.kt
+++ b/hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/OpenApiConfiguration.kt
@@ -1,17 +1,21 @@
 package uk.gov.justice.digital.hmpps.prisonersearch.search.config
 
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.databind.JavaType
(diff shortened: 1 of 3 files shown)


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/config/OpenApiConfiguration.kt
```diff
@@ -75,26 +71,6 @@ class OpenApiConfiguration(buildProperties: BuildProperties) {
...
     .addSecurityItem(SecurityRequirement().addList("prisoner-in-prison-search-role", listOf("read")))
     .addSecurityItem(SecurityRequirement().addList("prisoner-search--prisoner--ro", listOf("read")))
-
-  @Bean
-  fun openAPICustomiser(): OpenApiCustomizer = OpenApiCustomizer {
-    it.components.schemas.forEach { (_, schema: Schema<*>) ->
-      val properties = schema.properties ?: mutableMapOf()
-      for (propertyName in properties.keys) {
-        val propertySchema = properties[propertyName]!!
-        if (propertySchema is DateTimeSchema) {
```
petergphillips, line 85: not needed these days and actually stops dates being converted into `DateTime` by the OpenAPI generator

PR #622, opened 2025-08-15
SDIT-2993: 🐛 Fix model for openapi generation

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/attributesearch/api/Matcher.kt
```diff
@@ -19,7 +32,6 @@ import uk.gov.justice.digital.hmpps.prisonersearch.search.services.attributesear
   JsonSubTypes.Type(value = PncMatcher::class, name = "PNC"),
 )
 interface Matcher {
-  val type: String
```
petergphillips, line 22: this was effectively declared twice - once here and once in the `@JsonTypeInfo`.  This meant that the openapi generator got confused and didn't create the `type` in the children with an `override` modifier

PR #630, opened 2025-08-28
SDIT-2690: Be explicit in that only page and size are used by global search and flaws in paging

petergphillips:


andymarke (approved):
lgtm

PR #634, opened 2025-09-01
SDIT-3017: ✨ Add in response fields client mapper and use for restricted patients

petergphillips:


hmpps-prisoner-search/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/resource/RestrictedPatientsSearchResourceTest.kt
```diff
@@ -109,57 +109,6 @@ class RestrictedPatientsSearchResourceTest : AbstractSearchDataIntegrationTest()
         "/results/restrictedPatientsSearch/search_results_hosp_patient_one.json",
       )
     }
-
```
petergphillips, line 112: moved into nested class

PR #635, opened 2025-09-01
SDIT-3017: 🎨 Remove nested return and tidy

petergphillips:


hmpps-prisoner-search/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/search/services/RestrictedPatientSearchService.kt
```diff
@@ -61,30 +62,28 @@ class RestrictedPatientSearchService(
     pageSize: Int,
     pageOffset: Int,
     responseFields: List<String>? = null,
-    queryBuilder: (searchCriteria: RestrictedPatientSearchCriteria) -> BoolQueryBuilder?,
+    queryBuilder: (searchCriteria: RestrictedPatientSearchCriteria) -> BoolQueryBuilder,
   ): RestrictedPatientResult {
     responseFields?.run { responseFieldsValidator.validate(responseFields) }
     val query = queryBuilder(searchCriteria)
-    return query?.let {
```
petergphillips, line 68: `query` can never be null so didn't need `let`

PR #646, opened 2025-09-29
SDIT-3059: Disable cancel outdated as was cancelling all jobs

petergphillips:


.github/workflows/cancel_outdated_workflows.yml
```diff
@@ -1,8 +1,6 @@
 name: Cancel outdated waiting workflows
 on:
   workflow_dispatch:
-  schedule:
```
petergphillips, line 4: have removed the cron, but can still be run manually

mikehalmamoj (approved):

PR #652, opened 2025-10-06
SDIT-2804 Ignore expected errors for some batch jobs

mikehalmamoj:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/batch/BatchManager.kt
```diff
@@ -41,12 +43,25 @@ class BatchManager(
   @WithSpan
   fun runBatchJob(@SpanAttribute batchType: BatchType) = runBlocking {
     when (batchType) {
-      CHECK_INDEX_COMPLETE -> maintainIndexService.markIndexingComplete()
+      CHECK_INDEX_COMPLETE -> ignoreFailure<IndexException> { maintainIndexService.markIndexingComplete() }
       COMPARE_INDEX_SIZE -> compareIndexService.doIndexSizeCheck()
-      FULL_INDEX_REFRESH -> refreshIndexService.startIndexRefresh()
+      FULL_INDEX_REFRESH -> ignoreFailure<IndexException> { refreshIndexService.startIndexRefresh() }
       REMOVE_OLD_DIFFERENCES -> prisonerDifferencesService.deleteOldData()
     }
```
mikehalmamoj, line 50: When the app ends in error the Cronjob fails and is retried. Therefore we need to ignore expected exceptions for some batch jobs - e.g. don't end the check index complete batch job in error if an index isn't running, it's not an error.

PR #664, opened 2025-11-04
SDIT-3157 Add non-parole date to prisoner search

steverendell:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
```diff
@@ -327,6 +327,11 @@ class Prisoner : Diffable<Prisoner> {
   @DiffableProperty(DiffCategory.SENTENCE)
   var conditionalReleaseDate: LocalDate? = null
```
steverendell: i will uncomment DiffableProperty once it is in the index in prod

PR #752, opened 2026-04-15
SDIT-3668: 🐛 Fix api docs script as dependency order corrected

petergphillips:


check-api-docs.bash
```diff
@@ -33,16 +36,13 @@ if [[ "$OLD_VERSION" == "$NEW_VERSION" ]]; then
   exit 0
 fi
 
-# build the model again
-./gradlew "$MODEL_TASK"
-
 # and compare
 if ! diff -r "${WORKING_DIR}/build/model_copy" "${WORKING_DIR}/build/generated/$PROJECT" > "${WORKING_DIR}/build/api.diff"; then
   echo "Found differences between old ($OLD_VERSION) and new ($NEW_VERSION) models"
-  {
```
petergphillips, line 42: old version that I fixed in the other two projects

mikehalmamoj (approved):

PR #750, opened 2026-04-08
SDIT-3065 Readmission switch booking - failing edge case

steverendell:

PR #756, opened 2026-04-21
Revert "SDIT-3065 Readmission switch booking - failing edge case"

steverendell:
Reverts ministryofjustice/hmpps-prisoner-search#750


petergphillips (approved):
LGTM

PR #760, opened 2026-04-23
SDIT-3695 READMISSION_SWITCH_BOOKING not set for readmissions after a merge

steverendell:
use last movement creation timestamp to detect MERGE admission


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/nomis/OffenderBooking.kt
```diff
@@ -20,7 +20,7 @@ data class OffenderBooking(
   var lastMovementTypeCode: String? = null,
   var lastMovementReasonCode: String? = null,
   var lastMovementTime: LocalDateTime? = null,
-  // val lastMovementCreationTime: LocalDateTime? = null,
+  var lastMovementCreationTime: LocalDateTime? = null,
```
mikehalmamoj: I don't think this should be a var? If anything is immutable it's this. Looks like you only mutate it in tests. Not sure that's a good enough reason for var, is there some other way you can set up the tests?

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
```diff
@@ -158,7 +158,7 @@ private fun Prisoner.isAdmissionAssociatedWithAMerge(
   status == "ACTIVE IN" &&
   offenderBooking.identifiersForActiveOffender("MERGED")
     // check the merge is after the admission movement - or if there is no movement then check the merge happened in the last 90 minutes
-    ?.any { it.whenCreated > maxOf(offenderBooking.lastMovementTime ?: LocalDateTime.MIN, LocalDateTime.now().minusMinutes(90)) }
+    ?.any { it.whenCreated > maxOf(offenderBooking.lastMovementCreationTime ?: LocalDateTime.MIN, LocalDateTime.now().minusMinutes(90)) }
     ?: false
```
andymarke, line 162: I wonder if all this logic could massively be improved if we got the merge date time from the MERGE_TRANSACTION table - I feel that whole 90 minute stuff must have been before we even knew the merge time was recorded.

Feels like it could be something like 
1. Was movement created by merge (looking at audit module I guess)
2. Was it in-between the request time and audit time on the merge_transaction


Or something like that rather than - it would need a bit of an investigation, but I think the merge table would hold the time the period a merge was taken place

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/model/nomis/OffenderBooking.kt
steverendell: i just lazily followed the pattern of the other lastMovement* vars above, yes should really be a val

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/events/PrisonerMovementsEventService.kt
steverendell: Agree this sounds like a great idea (the 90m stuff is a bit arbitrary of course)

andymarke (approved):

PR #781, opened 2026-05-14
SDIT-3803: 🐛 Default alerts to empty list to prevent null pointers in clients

petergphillips:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
```diff
@@ -181,7 +181,7 @@ class Prisoner : Diffable<Prisoner> {
   @Field(type = FieldType.Nested, includeInParent = true)
   @Schema(description = "Alerts")
   @DiffableProperty(DiffCategory.ALERTS)
-  var alerts: List<PrisonerAlert>? = null
+  var alerts: List<PrisonerAlert>? = emptyList()
```
petergphillips, line 184: main change - default to empty list

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/repository/PrisonerRepository.kt
```diff
@@ -125,9 +125,7 @@ class PrisonerRepository(
     summary: PrisonerDocumentSummary,
   ): Boolean = doUpdate(
     prisonerNumber,
-    alerts?.let {
-      mapOf("alerts" to objectMapperWithNulls.convertValue(alerts))
-    } ?: mapOf("alerts" to null),
+    mapOf("alerts" to objectMapperWithNulls.convertValue(alerts)),
```
petergphillips, line 128: this was unnecessary as `convertValue` deals with `null` properly anyway

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/AlertsService.kt
```diff
@@ -46,7 +46,7 @@ class AlertsService(
         compareByDescending<Alert> { it.activeFrom }
           .thenBy { it.createdAt }, // Note this is Ascending
       )
-    }
+    } ?: emptyList()
```
petergphillips, line 49: default to empty list even if we get a `404` back from the alerts service

steverendell (approved):
lgtm

PR #783, opened 2026-05-18
SDIT-3803: 🐛 Second fix for defaulting alerts to empty list to prevent null pointers in clients

petergphillips:


steverendell (approved):
lgtm

PR #789, opened 2026-06-03
SDIT-3855: 👽️ Switch to new incentives prisoner endpoint

petergphillips:


hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/IncentivesService.kt
```diff
@@ -5,17 +5,23 @@ import org.springframework.retry.annotation.Retryable
...
 import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.prisonersearch.indexer.incentives.api.MaintainIncentiveReviewsApi
 import uk.gov.justice.digital.hmpps.prisonersearch.indexer.incentives.model.IncentiveReviewSummary
 
 @Service
 class IncentivesService(
-  private val incentivesWebClient: WebClient,
+  incentivesWebClient: WebClient,
 ) {
+  private val incentivesApi: MaintainIncentiveReviewsApi = MaintainIncentiveReviewsApi(incentivesWebClient)
```
petergphillips, line 17: switch to calling generated api

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/IncentivesService.kt
```diff
@@ -5,17 +5,23 @@ import org.springframework.retry.annotation.Retryable
...
+
   @Retryable(maxAttempts = 3, backoff = Backoff(delay = 100))
-  fun getCurrentIncentive(bookingId: Long): IncentiveReviewSummary? = incentivesWebClient.get().uri("/incentive-reviews/booking/{bookingId}?with-details=false", bookingId)
+  fun getCurrentIncentive(prisonNumber: String): IncentiveReviewSummary? = incentivesApi.prepare(
+    incentivesApi.getPrisonerIncentiveReviewHistoryRequestConfig(prisonerNumber = prisonNumber),
+  )
     .retrieve()
-    .bodyToMono(IncentiveReviewSummary::class.java)
+    .bodyToMono<IncentiveReviewSummary>()
```
petergphillips, line 24: intellij suggestion to use kotlin version of call

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -382,7 +382,9 @@ class PrisonerSynchroniserService(
     restrictedPatientService.getRestrictedPatient(it.offenderNo)
   }
 
-  private fun getIncentive(ob: OffenderBooking) = ob.bookingId?.let { b -> incentivesService.getCurrentIncentive(b) }
+  private fun getIncentive(ob: OffenderBooking) = ob.takeIf { ob.bookingId != null }?.let {
```
petergphillips, line 385: I've left the logic the same as it was, in that we only try to get the incentive if the prisoner has a booking (as there are some prisoners in NOMIS without any bookings)

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
```diff
@@ -148,7 +148,7 @@ class PrisonerSynchroniserService(
   internal fun reindexIncentive(prisonerNo: String, eventType: String) = prisonerRepository.getSummary(prisonerNo)
     ?.run {
       val bookingId = this.prisoner?.bookingId?.toLong() ?: throw PrisonerNotFoundException(prisonerNo)
```
petergphillips, line 150: I've left retrieving the `bookingId` here as it is used in the telemetry. The line of code seems a little bit odd though, since prisoners can exist without a booking. I'm guessing though that they won't have an incentive level unless they do have a booking so I think it is okay 🤞

steverendell: Interesting, the booking could be null if someone manually resynced a prisoner with no bookings, so that throw actually looks wrong in theory. Has probably never happened though 😄

hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/services/PrisonerSynchroniserService.kt
steverendell: Looks like the new endpoint returns a 404 if there are no incentives so the check is still needed (as a booking guarantees an incentive)

steverendell (approved):
lgtm

PR #793, opened 2026-06-11
SDIT-3887: ⬆️ Upgrade gradle plugin for spring boot 4.0.7

petergphillips:


hmpps-prisoner-search-indexer/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/MatchDomainEvents.kt
```diff
@@ -1,190 +0,0 @@
-package uk.gov.justice.digital.hmpps.prisonersearch.indexer
```
petergphillips, line 1: think this was accidentally committed @steverendell ?

karenmillermoj (approved):
lgtm

PR #798, opened 2026-06-28
MAPB-573 Add lastMovementDate to prisoner index

Mjwillis:
## What
Adds a new `lastMovementDate` (LocalDate) field to the prisoner search index, mapped from the `lastMovementTime` already returned by prison-api's `/api/prisoner-search/offenders/{offenderNo}` endpoint.

## Why
The locations-inside-prison API roll count (`getNumOvernights`) needs the last movement date to decide which CRT/TAP `ACTIVE OUT` prisoners are on overnight leave. Today it makes an extra prison-api call to learn this. Exposing the date on the prisoner index lets that consumer drop the extra round-trip.

## Changes
- `Prisoner` model: new `lastMovementDate` field (`@Field` date, diffable under LOCATION)
- `translator`: `lastMovementDate = ob.lastMovementTime?.toLocalDate()`
- Translator test for the mapping

## Note
A **full reindex** (dev → preprod → prod) is required after deploy to populate the field on existing documents.

### DIFF
--- a/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
+++ b/common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
@@ -144,7 +144,6 @@ class Prisoner : Diffable<Prisoner> {
 
   @Field(type = FieldType.Date, format = [DateFormat.date])
   @Schema(description = "Date of the last movement of the prisoner", example = "2023-05-01")
-  //@DiffableProperty(DiffCategory.LOCATION)
   var lastMovementDate: LocalDate? = null
 
   @Schema(description = "In/Out Status", example = "IN", allowableValues = ["IN", "OUT", "TRN"])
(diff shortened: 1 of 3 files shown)


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
```diff
@@ -142,6 +142,11 @@ class Prisoner : Diffable<Prisoner> {
   @DiffableProperty(DiffCategory.LOCATION)
   var lastMovementReasonCode: String? = null
 
+  @Field(type = FieldType.Date, format = [DateFormat.date])
+  @Schema(description = "Date of the last movement of the prisoner", example = "2023-05-01")
+  @DiffableProperty(DiffCategory.LOCATION)
```
steverendell: What I usually do is leave DiffableProperty commented out initially, that avoids accidently hitting the index refresh before the new field is populated and sending diff events for every prisoner

steverendell (approved):
lgtm

PR #803, opened 2026-07-10
SDIT-4027: 🐛 Fix smoke tests

petergphillips:


hmpps-prisoner-search-indexer/build.gradle.kts
```diff
@@ -24,10 +24,6 @@ dependencyCheck {
 testing {
   suites {
     register<JvmTestSuite>("testSmoke") {
-      dependencyCheck {
-        suppressionFiles.add("azure-dependency-check-suppress.xml")
```
petergphillips, line 28: accidentally left this in from previous suppressions

hmpps-prisoner-search-indexer/src/testSmoke/kotlin/uk/gov/justice/hmpps/offenderevents/smoketest/SmokeTestConfiguration.kt
```diff
@@ -20,4 +25,13 @@ class SmokeTestConfiguration(@Value("\${smoketest.endpoint.url}") private val sm
 
   @Bean
   fun smokeTestWebClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient = webClientBuilder.authorisedWebClient(authorizedClientManager, registrationId = "smoketest-service", url = smokeTestUrl)
+
+  @Bean
+  fun authorizedClientManager(
```
petergphillips, line 30: this class is importing the `OAuth2ClientAutoConfiguration` which in turn sets up a `DefaultOAuth2AuthorizedClientManager` which is only suitable for use within the context of an `HttpServletRequest`. Instead we need a `AuthorizedClientServiceOAuth2AuthorizedClientManager`

mikehalmamoj (approved):

PR #804, opened 2026-07-14
SDIT-3976: 📝 Remove required fields from swagger docs

petergphillips:


common/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/common/model/Prisoner.kt
```diff
@@ -358,7 +358,7 @@ class Prisoner : Diffable<Prisoner> {
   @DiffableProperty(DiffCategory.LOCATION)
   var locationDescription: String? = null
 
-  @Schema(required = true, description = "Indicates a restricted patient. Will never be null.", example = "true")
+  @Schema(description = "Indicates a restricted patient. Will never be null.", example = "true")
```
petergphillips, line 361: wasn't sure if I should change the swagger docs for all the other fields to include `Will never be null` or rather `Always returned if field specified` or something like that, but decided to leave it alone. Not sure if it holds true for all the other fields anyway...

mikehalmamoj (approved):

karenmillermoj (approved):
lgtm

